PR #36, opened 2025-07-14
Vector gate mlp

oli-clive-griffin:
Removes the existing `Gate` class (scalar affine gate), and adds a new `VectorGateMLP` which is a gate which reads from the vector-valued input to the layer being gated. For example, in in MLP up-proj layer, the input to the gate would be the residual stream.

Also:
- bumps torch to 2.6, due to [this](https://nvd.nist.gov/vuln/detail/CVE-2025-32434) vulnerability breaking huggingface `from_pretrained`. 
- fixes a bunch of type errors due to having to bump torch

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,14 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Python Debugger: Current File with Arguments",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${file}",
+            "console": "integratedTerminal",
+            "args": "${command:pickArgs}"
+        },
+        {
             "name": "Python: Attach",
             "type": "debugpy",
             "request": "attach",
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -115,8 +115,7 @@ class Config(BaseModel):
         description="Type of gate used to calculate the causal importance.",
     )
     gate_hidden_dims: list[NonNegativeInt] = Field(
-        ...,
-        description="Hidden dimensions for the gate used to calculate the causal importance"
+        ..., description="Hidden dimensions for the gate used to calculate the causal importance"
     )
     sigmoid_type: Literal["normal", "hard", "leaky_hard", "upper_leaky_hard", "swish_hard"] = Field(
         default="leaky_hard",
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -14,6 +14,7 @@ from pydantic import (
 )
 
 from spd.log import logger
+from spd.models.components import GateType
 from spd.spd_types import ModelPath, Probability
 
 
@@ -109,10 +110,13 @@ class Config(BaseModel):
         ...,
         description="Number of stochastic masks to sample when using stochastic recon losses",
     )
-    n_ci_mlp_neurons: NonNegativeInt = Field(
-        default=0,
-        description="Number of hidden neurons in the MLP used to calculate the causal importance."
-        "If 0, use a single-layer gate.",
+    gate_type: GateType = Field(
+        default="vector_mlp",
+        description="Type of gate used to calculate the causal importance.",
+    )
+    gate_hidden_dims: list[NonNegativeInt] = Field(
+        ...,
+        description="Hidden dimensions for the gate used to calculate the causal importance"
     )
     sigmoid_type: Literal["normal", "hard", "leaky_hard", "upper_leaky_hard", "swish_hard"] = Field(
         default="leaky_hard",
@@ -282,4 +286,5 @@ class Config(BaseModel):
             assert self.lr_exponential_halflife is not None, (
                 "lr_exponential_halflife must be set if lr_schedule is exponential"
             )
+
         return self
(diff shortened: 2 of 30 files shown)


.vscode/launch.json
```diff
@@ -4,6 +4,14 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        {
+            "name": "Python Debugger: Current File with Arguments",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${file}",
+            "console": "integratedTerminal",
+            "args": "${command:pickArgs}"
```
danbraunai, line 13: oh my lordy I had no idea this existed

spd/configs.py
```diff
@@ -109,10 +110,13 @@ class Config(BaseModel):
         ...,
         description="Number of stochastic masks to sample when using stochastic recon losses",
     )
-    n_ci_mlp_neurons: NonNegativeInt = Field(
-        default=0,
-        description="Number of hidden neurons in the MLP used to calculate the causal importance."
-        "If 0, use a single-layer gate.",
+    gate_type: GateType = Field(
```
danbraunai, line 113: Lee has brought up the good point that we should probably avoid the word "gate" everywhere. We're not actually learning the gates here, we're learning the ci function. The gates, if it's at all a reasonable word, would be the (stochastic) masks.

I know the word gate is very embedded throughout this codebase, but for now I'd like to avoid adding new instances of it, and would probably like to have a PR at some point which removes it everywhere.
I added #48 which asks to remove it throughout the codebase.

spd/models/components.py
```diff
@@ -8,58 +8,67 @@
 
 from spd.utils.module_utils import init_param_
 
+GateType = Literal["mlp", "vector_mlp"]
 
-class Gate(nn.Module):
-    """A gate that maps a single input to a single output."""
 
-    def __init__(self, C: int):
+class ParallelLinear(nn.Module):
```
danbraunai, line 14: This class is always called with a single output dimension in SPD, and I'm not sure if there are use cases where we might want a value other than 1.  I thus think it probably doesn't make sense to have it:
1. It's a little less clear to the reader when they read this class that Do is always 1.
2. I'm worried that operations become less efficient with the extra singleton dimension here. But I don't know, and it might be minimal.

spd/models/components.py
```diff
@@ -8,58 +8,67 @@
...
-        fan_val = 1  # Since each weight gets applied independently
-        init_param_(self.weight, fan_val=fan_val, nonlinearity="linear")
+        self.W_CDiDo = nn.Parameter(torch.empty(C, input_dim, output_dim))
+        self.bias_Do = nn.Parameter(torch.zeros(C, output_dim))
+        init_param_(self.W_CDiDo, fan_val=input_dim, nonlinearity="relu")
 
     @override
-    def forward(self, x: Float[Tensor, "... C"]) -> Float[Tensor, "... C"]:
-        return x * self.weight + self.bias
+    def forward(self, x_BxCDi: Tensor) -> Tensor:
```
danbraunai, line 24: Do you have a plan for switching to suffix types rather than jaxtypes? I think I'm onboard with suffix types after our discussions and more thought, but don't really like that this PR just mixes them both. To save time, I think it's OK to merge this in provided there is another PR that gets merged in the coming days that uses suffixes everywhere.

danbraunai (changes requested):
I'd like to see an evals run on all of tms and resid_mlp with a PR as big as this. It should get similar but not necessarily the same results. I've set one off [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250715_103042--VmlldzoxMzU5Njc3NQ==).

Apart from that, and a couple of comments, looks great.

spd/configs.py
oli-clive-griffin: I think this will make #39 tricky to merge if I change this now. Would you mind if I handled this after merging that?

spd/models/components.py
oli-clive-griffin: `Do` isn't always singleton: with multiple layers the first n-1 have scalar outputs. We Could have seperate implementations for the input (1 -> d in the GateMLP case), intermediate (d -> d), and output (d -> 1) layers but that seems a little overkill.

but yea I can test the performance to make sure this isn't slowing us down

spd/models/components.py
oli-clive-griffin: yea good point, I'll just keep to jax for now

.vscode/launch.json
oli-clive-griffin: ikr it's so great

spd/configs.py
danbraunai: Yep fair enough, someone can handle after (not on you).

oli-clive-griffin:
huh, weird, by reports I meant those from the evals runs you set off, agree they don't look great, will dig in and try and see whats going on

oli-clive-griffin:
found the problem, it was:
- the final layer wasn't init'd with a nonlinearity of "linear"
- the final layer had a gelu after it

oli-clive-griffin:
@danbraunai looking much better as far as I can tell

danbraunai (approved):
lgtm

PR #39, opened 2025-07-14
restructure handling of components and gates

oli-clive-griffin:
Restructures how we manage components and gates, relying much more on them attached to the `ComponentModel`.

Where we currently hold `components` separately to the model, this puts the replacement layer right next to the layer it replaces

This is a kinda artistic rendering:
```
ComponentModel
- model:
  - layer 14
    - mlp
      - up_proj
- components: {
  "model.layers.14.mlp.up_proj": ...
}
```

```
ComponentModel
- model:
  - layer 14
    - mlp
      - up_proj
        - original
        - replacement
```


spd/experiments/resid_mlp/resid_mlp_interp.py
```diff
@@ -648,11 +633,10 @@ def main():
         assert isinstance(target_model, ResidualMLP)
         n_layers = target_model.config.n_layers
 
-        components: dict[str, LinearComponent] = {
-            k.removeprefix("components.").replace("-", "."): v
-            for k, v in model.components.items()
-            if isinstance(v, LinearComponent)
-        }
+        components = {}
```
mivanit, line 636: why get rid of the type hints here?

oli-clive-griffin: nice catch, have improved

spd/models/components.py
```diff
@@ -161,10 +160,68 @@ def forward(self, x: Float[Tensor, "batch pos"]) -> Float[Tensor, "batch pos emb
...
+        super().__init__()
+        assert isinstance(original, nn.Linear) == isinstance(replacement, LinearComponent)
+        self.original = original
+        self.replacement = replacement
+
+        self.forward_mode: Literal["original"] | Literal["replacement"] | None = None
+        self.mask_BxC: Tensor | None = None
+        self.init_weights_()
+
+    def init_weights_(self) -> None:
```
danbraunai, line 188: I don't love that the init logic is here. This class feels like its job is just to choose which module to run through. Adding init logic seems out of place. I think one of the following might be nicer (roughly in order of niceness IMO):
1. Call a new init_weights function from both LinearComponent and EmbeddingComponent.
2. Create something like a Component class (from which LinearComponent and EmbeddingComponent inherits from), and that class this init function.
3. Just calling this init function after initialization of the class. Or maybe at the end of the ComponentModel init.

Thoughts?

spd/models/components.py
```diff
@@ -161,10 +160,68 @@ def forward(self, x: Float[Tensor, "batch pos"]) -> Float[Tensor, "batch pos emb
...
+            component_acts *= mask_BxC
 
         out = einops.einsum(
             component_acts, self.U, "batch pos C, ... C embedding_dim -> batch pos embedding_dim"
         )
         return out
+
+
+# TODO(oli) make this the only public class here
+class ReplacedComponent(nn.Module):
```
danbraunai, line 173: I don't like the name "ReplacedComponent". I prefer "ReplacementComponents" but even then I don't really like that either/

The class just chooses which of two nn.Modules to run. Maybe ModuleRouter? ModuleOrComponents (ew)? 

After saying this, I realise that I don't really like the singular in LinearComponent and EmbeddingComponent. Maybe they should just be pluralised. With singular, it makes me think that it's not necessarily the same components that we talk about in SPD, but just an object/module. If you feel inspired, pluralising these would be great (or make an issue).

spd/models/component_model.py
```diff
@@ -51,40 +53,58 @@ def __init__(
...
+        for component_name, component in replaced_components.items():
+            if gate_type == "mlp":
+                gates[component_name] = GateMLP(C=C, hidden_dims=gate_hidden_dims)
+            else:
+                input_dim = (
+                    component.original.weight.shape[1]
+                    if isinstance(component.original, nn.Linear)
+                    else component.original.num_embeddings
+                )
+                gates[component_name] = VectorGateMLP(
```
danbraunai, line 84: Just noting that we don't have tests for the embedding experiments yet. Were you able to run this PR with an embedding component?

spd/models/component_model.py
```diff
@@ -51,40 +53,58 @@ def __init__(
         self.model = base_model
         self.C = C
         self.pretrained_model_output_attr = pretrained_model_output_attr
-        self.components = self.create_target_components(
-            target_module_patterns=target_module_patterns, C=C
+
+        replaced_components = self.create_replaced_components(base_model, target_module_patterns, C)
+        self.replaced_components = replaced_components
+        self._replaced_components = nn.ModuleDict(
+            {k.replace(".", "-"): v for k, v in replaced_components.items()}
```
danbraunai, line 60: You no longer need this except to register modules (e.g. it's not called anywhere). If self.replaced_modules was a ModuleDict then you could get rid of it. A downside there is that if it was a ModuleDict then you have the annoying typing issues where it can't recognise the types of the values. Unsure how problematic this is. We could avoid this with a custom ModuleDict that does preserve type information of its values.

Thoughts?

spd/models/component_model.py
```diff
@@ -283,33 +284,57 @@ def from_pretrained(cls, path: ModelPath) -> tuple["ComponentModel", Config, Pat
         comp_model.load_state_dict(model_weights)
         return comp_model, config, out_dir
 
+    def calc_causal_importances(
+        self,
+        pre_weight_acts: dict[str, Tensor],
+        detach_inputs: bool = False,
+        sigmoid_type: SigmoidTypes = "leaky_hard",
+    ) -> tuple[dict[str, Float[Tensor, "... C"]], dict[str, Float[Tensor, "... C"]]]:
+        """Calculate component activations and causal importances in one pass to save memory.
```
danbraunai, line 293: Sorry this "component activations" part is unnecessary (I probably forgot to remove it fro man AI output). Can you remove?

spd/user_metrics_and_figs.py.example
```diff
@@ -1,82 +0,0 @@
-"""User-defined metrics and figures for SPD experiments.
```
danbraunai, line 1: This file is supposed to be in the repo. Check the Makefile.

tests/test_component_model.py
```diff
@@ -1,188 +1,140 @@
...
         super().__init__()
         self.linear1 = nn.Linear(10, 5, bias=True)
         self.linear2 = nn.Linear(5, 3, bias=False)
         self.embedding = nn.Embedding(100, 8)
-        self.other_layer = nn.ReLU()  # Non-target layer
+        self.other_layer = nn.ReLU()  # Non‑target layer (should never be wrapped)
 
     @override
-    def forward(self, x: Float[Tensor, "..."]):
+    def forward(self, x: Float[torch.Tensor, "... 10"]):  # noqa: D401,E501
```
danbraunai, line 23: nit: torch.Tensor -> Tensor

danbraunai (changes requested):
Love this one. Bunch of inline comments.

spd/models/components.py
danbraunai: My current priority order with no very strong preferences
- ModuleOrComponents (weird but explicit)
- ReplacedModule
- RouterModule

spd/models/component_model.py
```diff
@@ -51,40 +53,58 @@ def __init__(
         self.model = base_model
```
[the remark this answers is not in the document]
oli-clive-griffin: nice, will do

oli-clive-griffin:
for posterity, here's a run in the current state (in progress):

https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250716_165852--VmlldzoxMzYxNzIwNw==

spd/models/component_model.py
oli-clive-griffin: Yea, I've restructured this a bit, keen to talk through in person to make sure you think it works correctly

spd/models/component_model.py
oli-clive-griffin: yes

danbraunai (changes requested):
Great updates. Comments inline, just minor stuff.

spd/models/components.py
```diff
@@ -75,34 +76,87 @@ def forward(self, x: Float[Tensor, "... d_in"]) -> Float[Tensor, "... C"]:
...
+            C: Number of components
+            v_dim: Number of rows in the weight matrix
+            u_dim: Number of columns in the weight matrix
+        """
         super().__init__()
         self.C = C
-        self.d_in = d_in
-        self.d_out = d_out
+        self.V = nn.Parameter(torch.empty(v_dim, C))
+        self.U = nn.Parameter(torch.empty(C, u_dim))
```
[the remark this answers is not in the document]
danbraunai: It's very ingrained in everyone's minds that V multiplies the inputs. I think having the opposite anywhere in the code would be confuse (and anger :) ) people. So I think it'd be better to have:
- EmbeddingComponents be as they are, with V already multiplying the inputs
- LinearComponents be such that V multiplies the input. You still "mirror the shape", you just have V and U on opposite sides of the weight matrix in the Linear "weight".

spd/models/components.py
```diff
@@ -75,34 +76,87 @@ def forward(self, x: Float[Tensor, "... d_in"]) -> Float[Tensor, "... C"]:
         return x[..., 0]
 
 
-class LinearComponent(nn.Module):
-    """A linear transformation made from V and U matrices for SPD.
-
-    The weight matrix W is decomposed as W = U^T @ V^T, where V and U are learned parameters.
-    """
+class Components(ABC, nn.Module):
```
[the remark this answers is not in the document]
danbraunai: mmm yeah that is getting a little awkward. Guess it's fine.

oli-clive-griffin:
@claude thoughts?

danbraunai (approved):
Looks great. Feel free to merge after addressing the minor comments.

spd/models/component_model.py
```diff
@@ -117,88 +94,125 @@ def create_target_components(self, target_module_patterns: list[str], C: int) ->
...
-        for gate in self.gates.values():
-            gate.to(*args, **kwargs)
-        return self
+    @staticmethod
+    def create_components_or_modules(
+        target_model: nn.Module,
+        target_module_paths: list[str],
+        C: int,
+    ) -> dict[str, ComponentsOrModule]:
+        """Create target components for the model."""
```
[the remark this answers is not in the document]
oli-clive-griffin: nice callout, have made this comment way more comprehensive

PR #42, opened 2025-07-14
bernouli sampling

oclivegriffin:


spd/utils/component_utils.py
```diff
@@ -1,15 +1,51 @@
...
+        stochastic: bool,
+    ) -> Tensor:
+        ctx.save_for_backward(sigma)
+        z = torch.bernoulli(sigma) if stochastic else (sigma >= 0.5).to(sigma.dtype)
+
+        return z
+
+    @override
+    @staticmethod
+    def backward(  # pyright: ignore [reportIncompatibleMethodOverride]
```
Laplace418, line 33: Shouldn't this be multiplying  grad_outputs with (1-min)?

spd/configs.py
```diff
@@ -84,6 +84,31 @@ class LMTaskConfig(BaseModel):
...
+    )
+
+
+class BernoulliSampleConfig(BaseModel):
+    sample_type: Literal["bernoulli"] = Field(
+        default="bernoulli",
+        description="Type of sample to use for stochastic reconstruction",
+    )
+    min: float = Field(
+        default=0.0,
```
Laplace418, line 107: 1. I think the default should be 0.5
2. I think there should be a comment warning people against setting it to anything other than 0.5. Maybe even a line that gets printed to the log if min isn't 0.5. I suspect this'd it'd be easy for people play with the settings, find lower min values give great metrics across the board, and go with it.

causal_importances.png
Laplace418, line 1: Do we need these images?

PR #45, opened 2025-07-14
Allow running locally and improve logging interface

### LINKED ISSUE #6: Swap out prints for logs
**Migrated from:** goodfire-ai/spd-gf#60
**Original author:** @danbraunai

---

We let print statements creep in to scripts/run.py and probably elsewhere. Swap these out for logging like the rest of the repo for better control.

### LINKED ISSUE #10: Give option to run spd-run locally
**Migrated from:** goodfire-ai/spd-gf#54
**Original author:** @danbraunai

---

Currently, spd-run will always launch a slurm job. We should at least allow running locally if it's a single job being run. This way, a user wouldn't have to manually run the _decomposition.py script in order to run locally.

We should probably also allow running locally if multiple jobs by just running the sequentially.

mivanit:
This PR combines:
- https://github.com/goodfire-ai/spd-gf/pull/59
- https://github.com/goodfire-ai/spd-gf/pull/64

I talked with @nathu-goodfire and he asked me to finish the changes to his PR. Sorry for combining the two, but they both touch the same code.

### Add `--local` flag to spd-run for running experiments locally without SLURM submission.

- Move all tests of `spd.scripts.run` to `tests/scripts_run/`
- add new `tests/scripts_run/test_main.py`
  - **this was mostly written by Claude** and I have not examined it in great detail, but most of it seems reasonable
  - if a terser and more human-inspected set of tests makes more sense here, I can redo them

### replaced print statements with logger calls pretty much everywhere

- `spd.log.logger` is now a custom `_SPDLogger`
  - `_SPDLogger` provides the `.section()` and `.values()` methods to replace some commonly-used behavior that was previously done manually
- added `terse` format, which can optionally be enabled for console logging (omits timestamps/level for readability).
  - this can be enabled with `logger.set_format("console", "terse")`
  - `spd/scripts/run.py` provides a new cli option `log_format` which we can use to set this.

### minor changes
- added coverage reports generation with `make coverage`
  - these are currently placed in `docs/coverage`, which is gitignored -- not sure if we should ignore it
  - this requires the `pytest-cov` dependency
- added type hints in some areas for readability
- in places where we save figures and log the figure filename, we sometimes defined a figure name just once, and other times repeated the name in both the `savefig` and `print` calls. now, we always just declare a filename var first, and pass that to both `savefig` and `logger.info`
  - an aside -- should we have a `logger.savefig` which can save the figure, log that the figure was saved, and maybe keep track of all figures produced?

## Related Issues
- Closes #6 
- Closes #10 

## Motivation and Context
- Allow users to run experiments locally (instead of on a slurm cluster)
- Logging > Printing, especially if we are consistent about it
- `logger.section("thing")` > `print("="*50 + "\n" + "thing" + "\n" + "="*50)` if this is something we do often. 
  - in general, I think semantic logging is more useful: where we lean towards logging an object and the function call provides information about how to format things (since this lets us change things later)

## How Has This Been Tested?
- mock tests for job submission, check the args match (as per https://github.com/goodfire-ai/spd-gf/pull/59#pullrequestreview-3011953129)
- checked some things by hand

## Dependencies

- added `pytest-cov` to dev dependencies for coverage reports

only one: in `spd/experiments/tms/plotting.py`, `print_analysis_summary` is replaced with `get_analysis_summary` which gets a dict. the only call of that function (same file) has been adjusted appropriately.


oli-clive-griffin (commented):
Nice. I like using logger over print. In general I'd argue there's a lot of unnecessary type annotations and some unnecessary comments (see [this](https://www.youtube.com/watch?v=Bf7vDBBOBUA) video for imo a great argument against commenting lots).

Also while I like the `.values()` and `.section()` methods, I'm not sure how nicely they fit into the general logging framework and might be simpler to just use `.info()`? Don't have a strong take here though.

spd/scripts/run.py
```diff
@@ -330,12 +336,45 @@ def generate_commands(
...
     return commands
 
 
+def run_commands_locally(commands: list[str]) -> None:
+    """Execute commands locally in sequence.
+
+    Args:
+        commands: List of shell commands to execute
+    """
+    import shlex
```
danbraunai, line 354: I think I prefer just importing this at the top of the file with the rest of the imports. Especially because it's just a small library.

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
...
+    generate_commands,
+    generate_run_id,
+    main,
+    resolve_sweep_params_path,
+)
+
+
+def get_valid_tms_config():
+    """Get a valid TMS experiment config."""
+    return {
```
danbraunai, line 27: I think I'd prefer to just read one of our canonical configs (`spd/experiments/tms/tms_5-2_config.yaml`). Main reason is that it's very annoying to update all of configs in tests when something in the code changes.

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
...
+
+class TestCommandGeneration:
+    """Test command generation and utility functions.
+
+    Tests run_id generation, path resolution, and the integration of
+    generate_commands() with actual config loading and JSON serialization.
+    Grid combination logic is tested separately in test_grid_search.py.
+    """
+
+    def test_generate_run_id_format(self):
```
danbraunai, line 73: Overkill IMO, i'd remove.

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
...
+
+        assert config_json["wandb_project"] == "test-project"  # Should be overridden
+        assert config_json["C"] == 10
+        assert config_json["task_config"]["task_name"] == "tms"
+
+    @patch("spd.scripts.run.load_config")
+    @patch("spd.scripts.run.load_sweep_params")
+    def test_generate_commands_with_sweep(self, mock_load_sweep_params, mock_load_config):
+        """Test generate_commands function with sweep parameters."""
+        from spd.configs import Config
```
danbraunai, line 139: I'd just import this globally (it's used in the above fn too)

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
...
+
+        # Verify subprocess was called
+        assert mock_subprocess.call_count == 1
+
+        # Verify the command structure
+        args = mock_subprocess.call_args[0][0]
+        assert args[0] == "python"
+        assert "tms_decomposition.py" in args[1]
+
+        # Verify logger calls
```
danbraunai, line 203: I'd remove all logger verification in this and any other test it's in. We don't want to have to change the tests when changing the logs.

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
...
+        mock_load_sweep_params.return_value = {"C": {"values": [5, 10]}}
+        mock_subprocess.return_value = Mock(returncode=0)
+
+        # Should not raise an error
+        main(experiments="tms_5-2-id", sweep=True, local=True)
+
+        assert mock_subprocess.call_count == 2
+
+
+class TestLoggerOutput:
```
danbraunai, line 481: For same reason as above, I'd remove this. Don't care about logger calls.

tests/scripts_run/test_main.py
```diff
@@ -0,0 +1,774 @@
+"""Tests for the main() function and end-to-end execution workflows in spd/scripts/run.py.
+
+This file focuses on testing the high-level integration behavior of the main() function,
+including local execution, SLURM submission, and end-to-end parameter sweep workflows.
+
+Lower-level functions like generate_grid_combinations and load_sweep_params are tested
+in test_grid_search.py and test_run_sweep_params.py respectively.
+"""
+
+# pyright: reportUnknownParameterType=false, reportMissingParameterType=false, reportUnusedParameter=false
```
danbraunai, line 10: I made some minor comments inline. But after thinking and discussing with Oli, I think we should just avoid nearly all integration/end-to-end tests. It's just very annoying to change tests when you change the code, and the cost of things breaking that aren't caught in the tests is just very small, because the user will get an error and we can fix it straight away.

I do think that the cost of the tests is rapidly reducing as AI improves, but I still don't think it's enough to justify.

I do like the idea of simple happy-path tests which runs a couple of iterations of SPD, like we have in tests/test_resid_mlp.py and tests/test_tms.py. It'd be nice to do a similar one for spd-run --local. But I'd prefer not having all the tests in this file.

danbraunai (changes requested):
Love the --local and logging setup. Added to Oli's comments that mostly relate to code style, which we should have a PR about (#47). Main thing is avoiding integration/end-to-end tests and type hints.

spd/experiments/lm/play.py
```diff
@@ -76,21 +77,30 @@
...
-print("inputs_shape", input_ids.shape)
-print("logits", logits)
-print("logits shape", logits.shape)
+logger.values(
+    dict(
+        inputs_shape=input_ids.shape,
+        logits=logits,
+        logits_shape=logits.shape,
+    )
+)
```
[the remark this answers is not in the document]
mivanit: Thanks! Is this something we should stick to in general? I generally use `dict(key="val")` syntax when the keys correspond to variable names or other identifiers that should not just be any string -- i.e. when we want to put keys in a filename.

spd/scripts/run.py
```diff
@@ -374,12 +425,16 @@ def main(
         # Use custom W&B project
         spd-run --experiments tms_5-2 --project my-spd-project
     """
+    # Set logger format
+    logger.set_format("console", log_format)
```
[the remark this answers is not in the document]
mivanit: Re this and the [other comment on `set_format`](https://github.com/goodfire-ai/spd/pull/45#discussion_r2206754714):

I didn't want to change the default logging style -- unlikely that something depends on it, but I was trying to be careful. It would be much easier to just set the default `logging_config` to have

```python
"console": {
  "class": "logging.StreamHandler",
  "formatter": "terse", # instead of default
  "level": "INFO",
},
```

but we might sometimes want to do verbose logging in console -- not sure.

The reason we have to create this `set_format` function is because `setup_logger` is called on the load of `spd/log.py` so there's no way to modify the config before the logger is created.

spd/experiments/lm/component_viz.py
```diff
@@ -50,16 +52,20 @@ def main(path: ModelPath) -> None:
...
+            "mean_n_active_components_per_token": str(mean_n_active_components_per_token),
+            "mean_component_activation_counts": str(mean_component_activation_counts),
+        }
+    )
     fig = plot_mean_component_activation_counts(
         mean_component_activation_counts=mean_component_activation_counts,
     )
     # Save the entire figure once
-    save_path = out_dir / "modules_mean_component_activation_counts.png"
+    save_path: Path = out_dir / "modules_mean_component_activation_counts.png"
```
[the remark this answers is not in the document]
mivanit: I usually err on the side of "type hints everywhere" and kind of absentmindedly add them whenever I'm reading a piece of code -- sorry!

I'm happy to stick to the style guide now that we have one, but I've found previously that not having type hints *sometimes* leads to type hints being absent even when they are really useful and needed, since "deciding whether to add a type hint" is mental overhead and it becomes easier to just not add any.

tests/scripts_run/test_main.py
danbraunai: Sorry, bad wording by me. I did mean "Remove all the tests". My concern is that these tests touch too many specific functions, such that if you change the function names or structure or paths or whatever, you have to change the tests. What I think would be best for testing the spd-run script is a couple of integration tests which don't mock out stuff except for the final output of the command that will be run (or something near the final output). E.g. something like this
```
@patch(create_slurm_array_script)
def test_spd_run_not_local_no_sweep():
    # 1. Call spd.scripts.run.main with the standard arguments we use in our toy models (you might want a couple of tests to test variations).
    # 2. assert that create_slurm_array_script is called with the args we would expect


@patch(spd.scripts.run.subprocess.run)
def test_spd_run_local_no_sweep():
    # 1. Call spd.scripts.run.main with the standard arguments we use in our toy models (you might want a couple of tests to test variations).
    # 2. assert that the call to subprocess.run is as you would expect
```

danbraunai: I might not even check that ALL of the arugments to create_slurm_array_script are as they should be, maybe just the script_path. Otherwise we fall into the same issue where we've hardcoded a lot of specific names/functionality in this test.

mivanit: added options to `spd/scrips/run.py:main()` for:
- `use_wandb: bool = True` -- disables completely any wandb behavior
- `override_branch: str|None = None` -- default behavior when `None` is to create snapshots like before (only relevant when submitting to slurm). if set to a string, will check out that branch in the slurm script. disabling creation of snapshots would be messy -- passing `create_slurm_array_script(..., snapshot_branch=None)` from `main()` results in `create_slurm_array_script` doing the snapshot creation (side note: why duplicate this behavior?). so instead, we'll just pass the `dev` branch in the tests

danbraunai: > (side note: why duplicate this behavior?)
Yeah the git snapshot shouldn't be duplicated, you can fix that here (or add an issue). I'd want to have a better look at this override_branch argument tomorrow. Bit confused about when someone would want to specify their own string for it. Also the slurm script will fail if that branch doesn't exist. My naive intuition is just that you have a `create_snapshot: bool` but I'll have to understand why you said that would be messy.

> It's not clear to me what we can actually check about the script_path passed to create_slurm_array_script. It's defined as

Yeah i definitely wouldn't check that. We only really care about the "commands" argument. I might just check that one.

Thanks for going back and forth a lot on this one. Should be almost there :).

mivanit: removed the snapshot creation, but kept the ability to explicitly specify the branch. I think we could now rewrite it to only true or false for creating a snapshot branch, and use the current branch if false. other changes to the tests should also be good to go?

mivanit:
@danbraunai had to merge dev into this branch, and the merge for `spd/scripts/run.py` ended up being pretty messy. would appreciate if you could sanity check what I did there.

danbraunai:
@claude could you review this PR? I'm interested in high level questions about the structure of the new features and any ways that it could be improved. I want critical comments, do feel comfortable to give me the hard truths.

danbraunai:
@mivanit this was a nice claude review, I think the prompt was pretty good. Comments on some of the things it picked up:
1. custom logger design. I don't mind this too much as this is a pretty standalone project that won't be tightly integrated with others.
2. I like the Idea of the ExecutionStrategy(Protocol). But I wouldn't bother changing the current setup, since it's reasonable and it works.
3. spd/log.py L:68 "Issue: This will fail in environments without a terminal (CI, background processes). Need fallback handling." sounds like it might be an issue. Might be good to check/fix that.

Can merge if 3 is a simple fix and the other concerns in the previous comment are addressed without non-trivial changes.

mivanit:
> @mivanit this was a nice claude review, I think the prompt was pretty good. Comments on some of the things it picked up:
> 
>     1. custom logger design. I don't mind this too much as this is a pretty standalone project that won't be tightly integrated with others.
> 
>     2. I like the Idea of the ExecutionStrategy(Protocol). But I wouldn't bother changing the current setup, since it's reasonable and it works.
> 
>     3. spd/log.py L:68 "Issue: This will fail in environments without a terminal (CI, background processes). Need fallback handling." sounds like it might be an issue. Might be good to check/fix that.
> 
> 
> Can merge if 3 is a simple fix and the other concerns in the previous comment are addressed without non-trivial changes.

1. agree, I think vendor lock in to our own repo is fine
2. after taking a look at claude's suggestion, I actually kind of like the idea of an execution strategy. Might be best left for a future PR though
3. claude made a mistake here -- the first (optional) argument to `shutil.get_terminal_size` is the fallback size. Made it passed by keyword to make that more clear.

mivanit:
actually on further thought, I'm less convinced that actually cutting `main()` at running the commands is a good idea. I do however agree that the `main()` function is fairly verbose, so:

- getting the experiment names and validating them is moved to `get_experiments` (see [af6de44](https://github.com/goodfire-ai/spd/pull/45/commits/af6de4431075dc839fe2ebdfcccdce967407f451))
- the wandb setup and logging about it, which doesn't need to return anything at all, is moved to `_wandb_setup` (see [8aba920](https://github.com/goodfire-ai/spd/pull/45/commits/8aba920a3d549b6e9a6a70c73328c798cf69b853))

both of these commits are self-contained so we can revert either -- of the two, I'm less sure about factoring out the wandb stuff since it's a bit opaque, but all we were doing is calling already-opaque functions in main() directly, in a way that made the scope of some variables mildly confusing.

as a side note -- would it not be cleaner for `generate_commands` to return `list[list[str]]` instead of `list[str]` since we split the commands up again in `run_commands_locally`?

PR #47, opened 2025-07-15
Add STYLE.md for code style

oli-clive-griffin:
straighforward - adds a code style guide. My justification for this being useful is that it prevents back and forth about style which is somewhat subjective. In general, agreeing on and obeying a consistent coding style is much more beneficial than picking "the right" coding style.

Very open to suggestions here

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,15 +1,21 @@
 # Code Style Guide
 
 TLDR:
-- simple, straightforward code
-- fail fast - no defensive programming
-- use einops, jaxtyping, and shape assertions for tensor shape clarity
+- prioritise simple, straightforward code. Our users are researchers, often with little coding experience.
+- safety: use types, einops, jaxtyping, and liberal assertions.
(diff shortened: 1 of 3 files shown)


oli-clive-griffin:
@danbraunai what do you think?

STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+
+If there's an assumption you're making while writing code, assert it.
+- If you were right, then it won't matter
+- If you were wrong, then the code **should** fail.
+
+```python
+# BAD - silently handles unexpected state
+def process_activations(acts):
+    if acts is None:
+        return torch.zeros(hidden_size)  # Hides the problem
```
danbraunai, line 24: I can't see someone writing this function, but don't have an immediately better option. This might mean that providing examples here might not be worth it, and leaving just the bullet points above is fine. But if you have seen a lot of things like this, then I'm OK keeping it.

danbraunai (approved):
Minor comment but the rest looks good. Thanks!

PR #51, opened 2025-07-16
add extra component functions

casperlchristensen:
## Description

Adds a couple of useful functions to the component model. Specifically:

1. `forward_with_components_and_pre_forward_cache_hooks` allows us to obtain the pre-forward activations on the model with replaced components
2. `forward_with_components_and_post_forward_hooks` allows us to obtain the component outputs for each decomposed layer
3. `model_method_with_components` allows us to call any method of the underlying model, but with the components replaced. A concrete example is if we decompose a Transformer-model that implements e.g. `get_attention_map(...)`

## Motivation and Context
If there are any intermediate outputs in a decomposed model that are useful for analysis, this allows us to obtain those. For `model_method_with_components` it also allows us to not have to recompute useful metrics. For instance, if we wanted to compute the attention-maps given by a decomposed QK-circuit, we would normally have to:

1. Obtain the Q and K outputs for the decomposed model
2. Manually compute the attention, making sure we've applied the same keywords, scaling etc.

## How Has This Been Tested?

I've used it extensively for analysis locally -- I am quite certain it works.

Nope!

### DIFF
--- a/spd/models/component_model.py
+++ b/spd/models/component_model.py
@@ -206,6 +206,60 @@ class ComponentModel(nn.Module):
             for handle in handles:
                 handle.remove()
 
+    def forward_with_components_and_pre_forward_cache_hooks(
+        self,
+        *args: Any,
+        components: dict[str, LinearComponent | EmbeddingComponent],
+        masks: dict[str, Float[Tensor, "... C"]] | None = None,
+        **kwargs: Any,
+    ) -> tuple[Any, dict[str, Tensor]]:
+        """Forward pass with temporary component replacements and pre-forward cache hooks.
+
+        Args:
+            components: Dictionary mapping component names to components
+            masks: Optional dictionary mapping component names to masks
+        """
+        with self._replaced_modules(components, masks):
+            return self.forward_with_pre_forward_cache_hooks(
+                *args, module_names=list(components.keys()), **kwargs
+            )
+
+    def forward_with_components_and_post_forward_hooks(
+        self,
+        *args: Any,
+        components: dict[str, LinearComponent | EmbeddingComponent],
+        masks: dict[str, Float[Tensor, "... C"]] | None = None,
+        **kwargs: Any,
+    ) -> tuple[Any, dict[str, Tensor]]:
+        """Forward pass with temporary component replacements and post-forward hooks.
+
+        Args:
+            components: Dictionary mapping component names to components
+            masks: Optional dictionary mapping component names to masks
+        """
+        with self._replaced_modules(components, masks):
+            return self.forward_with_post_forward_cache_hooks(
+                *args, **kwargs, module_names=list(components.keys())
+            )
+
+    def model_method_with_components(
+        self,
+        *args: Any,
+        method_name: str,
+        components: dict[str, LinearComponent | EmbeddingComponent],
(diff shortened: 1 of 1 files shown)


spd/models/component_model.py
```diff
@@ -206,6 +206,60 @@ def cache_hook(_: nn.Module, input: tuple[Tensor, ...], param_name: str) -> None
...
+    def model_method_with_components(
+        self,
+        *args: Any,
+        method_name: str,
+        components: dict[str, LinearComponent | EmbeddingComponent],
+        masks: dict[str, Float[Tensor, "... C"]] | None = None,
+        **kwargs: Any,
+    ) -> Any:
+        """
+        Calls any method of the model with the given components.
```
danbraunai, line 254:
```suggestion
        Calls any method of the target model on a forward pass that routes through the components.
```

spd/models/component_model.py
```diff
@@ -206,6 +206,60 @@ def cache_hook(_: nn.Module, input: tuple[Tensor, ...], param_name: str) -> None
...
+        Args:
+            components: Dictionary mapping component names to components
+            masks: Optional dictionary mapping component names to masks
+        """
+        with self._replaced_modules(components, masks):
+            return self.forward_with_post_forward_cache_hooks(
+                *args, **kwargs, module_names=list(components.keys())
+            )
+
+    def model_method_with_components(
```
danbraunai, line 245:
```suggestion
    def target_method_with_components(
```

spd/models/component_model.py
```diff
@@ -206,6 +206,60 @@ def cache_hook(_: nn.Module, input: tuple[Tensor, ...], param_name: str) -> None
...
+        **kwargs: Any,
+    ) -> tuple[Any, dict[str, Tensor]]:
+        """Forward pass with temporary component replacements and post-forward hooks.
+
+        Args:
+            components: Dictionary mapping component names to components
+            masks: Optional dictionary mapping component names to masks
+        """
+        with self._replaced_modules(components, masks):
+            return self.forward_with_post_forward_cache_hooks(
```
danbraunai, line 241: I think this method relies on your other PR?

danbraunai (changes requested):
These seem useful, thanks for adding them! Though I think I'd prefer to have a general `forward_with_cache_hooks` method which takes arguments `pre_or_post: Literal["pre", "post"]` or `hook_position: Literal["pre", "post"]` and `forward_mode: Literal["original", "components"]`. I took this last one from https://github.com/goodfire-ai/spd/pull/39 which changes the structure of ComponentModel a lot and which we will probably merge in first.

Thoughts on unifying all of these?

casperlchristensen:
> These seem useful, thanks for adding them! Though I think I'd prefer to have a general `forward_with_cache_hooks` method which takes arguments `pre_or_post: Literal["pre", "post"]` or `hook_position: Literal["pre", "post"]` and `forward_mode: Literal["original", "components"]`. I took this last one from #39 which changes the structure of ComponentModel a lot and which we will probably merge in first.
> 
> Thoughts on unifying all of these?

Yeah makes perfect sense to me. I'll re-factor this a bit once the other component-related PR is merged then.

oli-clive-griffin:
@casperlchristensen @danbraunai seems like this is redundant after https://github.com/goodfire-ai/spd/pull/179?

PR #55, opened 2025-07-16
makefile improvements

### LINKED ISSUE #5: Fix uv issues
**Migrated from:** goodfire-ai/spd-gf#62
**Original author:** @mivanit

---

The recipes which call BasedPyright.........................................(no files to check)Skipped
Ruff lint............................................(no files to check)Skipped
Ruff format..........................................(no files to check)Skipped rely on either:
- BasedPyright.........................................(no files to check)Skipped
Ruff lint............................................(no files to check)Skipped
Ruff format..........................................(no files to check)Skipped being installed globally
- the environment being active

this is inconsistent with Provide a command or script to invoke with `uv run <command>` or `uv run <script>.py`.

The following commands are available in the environment:

- _vscode_copy_basedpyright-langserver
- basedpyright
- basedpyright-langserver
- convert-caffe2-to-onnx
- convert-onnx-to-caffe2
- datasets-cli
- debugpy
- debugpy-adapter
- dotenv
- f2py
- fonttools
- get_gprof
- get_objgraph
- huggingface-cli
- identify-cli
- ipython
- ipython3
- isympy
- jsonschema
- jupyter
- jupyter-kernel
- jupyter-kernelspec
- jupyter-migrate
- jupyter-run
- jupyter-troubleshoot
- nodeenv
- normalizer
- numpy-config
- pre-commit
- proton
- proton-viewer
- py.test
- pyftmerge
- pyftsubset
- pygmentize
- pytest
- python
- python3
- python3.12
- ruff
- spd-run
- streamlit
- streamlit.cmd
- tiny-agents
- torchfrtrace
- torchrun
- tqdm
- transformers
- transformers-cli
- ttx
- undill
- virtualenv
- wandb
- watchmedo
- wb

See `uv run --help` for more information. being used for the  and  recipes.

mivanit:
## Description
We want a single source of truth for common repo tasks, and having it be the makefile is nice. `make` is installed by default pretty much everywhere, and using it instead of `pre-commit` as the source of truth means running tasks does not depend on having the virtual environment set up properly.

## Related Issue
Closes #5 

Specifically, see https://github.com/goodfire-ai/spd/issues/5#issuecomment-3079215722

## Motivation and Context
- `make` recipes are now the source of truth for themselves and for `pre-commit`
- template copying is cleaner

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+RUN = uv run
+
 .PHONY: install
 install:
 	uv sync --no-dev
@@ -13,7 +15,7 @@ install:
 .PHONY: install-dev
 install-dev:
 	uv sync
-	pre-commit install
+	$(RUN) pre-commit install
 	@if [ ! -f spd/user_metrics_and_figs.py ]; then \
 		cp spd/user_metrics_and_figs.py.example spd/user_metrics_and_figs.py; \
 		echo "Created spd/user_metrics_and_figs.py from template"; \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,19 +1,15 @@
+# setup
 .PHONY: install
-install:
+install: copy-templates
```
danbraunai, line 3: I think after the removal of "copy-templates" and leaving the more verbose code copying, this PR will be good to merge.

danbraunai (changes requested):

Makefile
mivanit: The `copy-templates` recipe uses the more verbose copying, the only thing I changed is that the two install recipes now call `copy-templates` instead of duplicating that code. I can also change it back, but I think the makefile syntax for dependencies on other recipes (used here) is fairly straightfoward -- at least compared to dependencies on files when `.PHONY: <recipe_name>` is missing, which is I switched it to earlier, and what I thought the issue was. I can add some comments explaining things.

Current code, diff is a bit hard to read:
```make
.PHONY: install
install: copy-templates
	uv sync --no-dev

.PHONY: install-dev
install-dev: copy-templates
	uv sync
	pre-commit install

.PHONY: copy-templates
copy-templates:
	@if [ ! -f spd/user_metrics_and_figs.py ]; then \
		cp spd/user_metrics_and_figs.py.example spd/user_metrics_and_figs.py; \
		echo "Created spd/user_metrics_and_figs.py from template"; \
	fi
	@if [ ! -f spd/scripts/sweep_params.yaml ]; then \
		cp spd/scripts/sweep_params.yaml.example spd/scripts/sweep_params.yaml; \
		echo "Created spd/scripts/sweep_params.yaml from template"; \
	fi
```

danbraunai (approved):
I misread the previous changes. Looks good to merge now, feel free to do it whenever!

PR #63, opened 2025-07-18
Fix user metrics example file not being checked

mivanit:
# Description

the file `spd/user_metrics_and_figs.py.example` is never checked by `ruff` or `basedpyright`, as far as I can tell, because it is not a `.py` file.

This PR fixes that:

- before running `make format`, `make type`, or `make check`, we copy that file with the extension `.py` to `tests/.temp`
  - only then do we run the checks
- we can't `.gitignore` the `tests/.temp` dir, because then `ruff` and `basedpyright` won't check it
  - so, as part of the `pre-commit` pipeline, we remove that dir using the recipe `make temp-user-metrics-clean`

# Dependencies

Depends on #45, since that also modifies the makefile

# How was this tested

putting various errors in the user metrics example is uncaught without this, but is caught when we add this to the makefile/pre-commit


danbraunai:
We no longer have the spd/user_metrics_and_figs.py.example file

PR #65, opened 2025-07-19
Components Restructure - Second Try

oli-clive-griffin:
## Description
I broke dev with #39. Noticed [here](https://opensourcemechanistic.slack.com/archives/C08N7E5KNG7/p1752886755751499). I've found what was wrong. I had the freezing of parameters wrong: When we froze the target model params, we were inadvertently freezing the `ComponentsOrModule`s. Now we freeze the `target_model` **before** extending it with these. I've also added a test for this.

I've created a temporary draft PR [here](https://github.com/goodfire-ai/spd/pull/66) to display the difference between this PR and the last problematic one (#39)

## Related Issue
N/A

## Motivation and Context
see above

## How Has This Been Tested?
Added `test_correct_parameter_requires_grad`

N/A


oli-clive-griffin:
Demo run: https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250719_141112--VmlldzoxMzY1NDczNQ==

oli-clive-griffin:
I've created a temporary draft PR [here](https://github.com/goodfire-ai/spd/pull/66) to display the difference between this PR and the last problematic one (#39)

oli-clive-griffin:
hm ok, did a run on this commit and everything worked but tms 5-2, doing a seed sweep here:
https://wandb.ai/goodfire/spd?nw=thr378yuyvr

spd/models/component_model.py
```diff
@@ -117,92 +98,167 @@ def create_target_components(self, target_module_patterns: list[str], C: int) ->
...
             masks: Optional dictionary mapping component names to masks
         """
-        with self._replaced_modules(components, masks):
+        with self._replaced_modules(masks):
             return self(*args, **kwargs)
 
-    def forward_with_pre_forward_cache_hooks(
-        self, *args: Any, module_names: list[str], **kwargs: Any
+    def forward_with_component_pre_forward_cache_hooks(
+        self, *args: Any, **kwargs: Any
```
danbraunai, line 261: This name change is confusing. The new name implies that you're doing forward passes with the components, but you're actually just doing a forward pass on the target model. Why not leave it as is?

Note that Casper will probably expand on this functionality to allow all combinations of possible caching that the user might want https://github.com/goodfire-ai/spd/pull/51#issuecomment-3089306627.

danbraunai (approved):
I tried to find what the actual difference with this PR compared to the last one was, but couldn't https://github.com/goodfire-ai/spd/compare/d5fb7ffa653d3b54898e7881ab506567e97bc69f...oli/ioc. You mentioned something about parameter freezing?

- `dev`: 20 seeds of tms_5-2 has 2/20 failed (seeds 0-9 [here](https://wandb.ai/goodfire/spd?nw=lu99lq0om0x), seeds 10-19 [here](https://wandb.ai/goodfire/spd?nw=drlkx4j239m)) 
- `oli/ioc`: 20 seeds of tms_5-2 also has 2/20 failed (seeds 0-19 [here](https://wandb.ai/goodfire/spd/panel/7lnelfsjn?nw=ntzy11xzexi))

So we're good in that respect.

Happy to merge but would want:
1. Better PR description indicating what was actually different (copy of code)
2. Address the comment in the review about the new name `forward_with_component_pre_forward_cache_hooks`

spd/models/component_model.py
oli-clive-griffin: Fair. the change here is that I just realised we only ever cache *all* components' pre forward, so just removed the ability to pick.

If we want to allow more dynamic options, I can just revert this and leave as is.

oli-clive-griffin: have done this

oli-clive-griffin:
@danbraunai did you see the PR: https://github.com/goodfire-ai/spd/pull/66

danbraunai:
@oli-clive-griffin 

> @danbraunai did you see the PR: https://github.com/goodfire-ai/spd/pull/66
Yep OK ty. 

Fix looks good! Fine to merge.

Though one thing I realise I don't like about this change is that when you do component_model.target_model you don't actually get the target_model. Right now, if you call `target_model(data)` you'll get an error because forward_mode on the ComponentsOrModule objects will presumably be set to the default of None. Maybe we want to make the default "original" to prevent this? A bit less safety when doing that though. Even if we did that, it's still a bit weird that the target_model isn't actually the target_model though.

I'm now thinking we should change the attribute name from `target_model` to just `model` to prevent confusion here since it's no longer the target_model after initialization. Thoughts? (I know I previously suggested changing it from base_model to target_model, but didn't realise this issue at the time.)

oli-clive-griffin:
run [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250721_165414--VmlldzoxMzY3NjQ0Mw==) 🤞

oli-clive-griffin:
Update for readers: we spoke in person and decided `patched_module` works well.

changes since Dan's last comment:
- made ComponentModel.patched_model generic. A gotcha is that `.patched_model`'s patches modules are typed wrong. there's casts in the codebase for this
- tidied our usage of `.eval` and `.require_grad_` such that the target model always goes into the `ComponentModel` with all params non-trainable. We also assert this on `ComponentModel.__init__`
- Fixed `ComponentModel.from_pretrained`. was broken

spd/core_metrics_and_figs.py
```diff
@@ -0,0 +1,170 @@
...
+
+try:
+    from spd.user_metrics_and_figs import compute_user_metrics, create_user_figures
+except ImportError:
+    compute_user_metrics = None
+    create_user_figures = None
+
+
+def create_metrics(
+    model: ComponentModel[Any],
```
danbraunai, line 38: You've used nn.Module in the generic bound down below, do you want to use that by default?

spd/experiments/lm/app.py
```diff
@@ -35,12 +33,10 @@
 # -----------------------------------------------------------
 @dataclass(frozen=True)
 class AppData:
-    model: ComponentModel
+    model: ComponentModel[Any]
```
danbraunai, line 36: Same qn about Any vs nn.Module

spd/losses.py
```diff
@@ -126,10 +128,9 @@ def calc_importance_minimality_loss(
 
 
 def calc_masked_recon_layerwise_loss(
-    model: ComponentModel,
+    model: ComponentModel[Any],
```
danbraunai, line 131: same qn about Any vs nn.Module.

danbraunai: I realise now that this is everywhere, so do a ctrl-f if changing.

danbraunai: Oh, Any does map to nn.Module because the class definition is `ComponentModel[T: nn.Module]`? That's pretty gross if so.

danbraunai (approved):
Looks good. Though I don't know about `ComponentModel[T: nn.Module]`. What's the benefit of it if you have to do an isinstance following an instantiation of it anyway?

I can imagine it being useful inside the ComponentModel class if it gets used a lot. But I'm not sure the current use there justifies it. Outside of the class, it seems like you don't really care what time type the wrapped module is unless you're going to use an attribute of it, in which case you have to do an isinstance anyway?

But maybe I'm missing something here.

I see the value in DataLoader where it tells you the return type of the loader.

Happy for you to remove the type and merge or to argue for it (which very well might change my mind).

danbraunai (approved):
Nice. Good to merge after addressing the minor comments below (unless big changes).

CLAUDE.md
```diff
@@ -65,7 +65,8 @@ Both installation commands automatically create `spd/user_metrics_and_figs.py` f
 - `spd/models/component_model.py` - Core ComponentModel that wraps target models
 - `spd/models/components.py` - Component types (LinearComponent, EmbeddingComponent, etc.)
 - `spd/losses.py` - SPD loss functions (faithfulness, reconstruction, importance minimality)
-- `spd/user_metrics_and_figs.py` - User-defined metrics and visualizations (created from template)
+- `spd/metrics.py` - SPD metrics (faithfulness, reconstruction, importance minimality)
+- `spd/figures.py` - SPD figures (faithfulness, reconstruction, importance minimality)
```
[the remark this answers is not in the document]
oli-clive-griffin: ah, I'd meant to fix this in #71 but forgot to push the commit before merging. Having made a PR here: https://github.com/goodfire-ai/spd/pull/75/files

PR #66, opened 2025-07-19
temp: PR to display restrcture fix

oli-clive-griffin:


spd/models/component_model.py
```diff
@@ -152,15 +153,19 @@ def create_components_or_modules(
 
             if isinstance(module, nn.Linear):
                 d_out, d_in = module.weight.shape
-                component = LinearComponents(C=C, d_in=d_in, d_out=d_out, bias=module.bias)
+                component = LinearComponents(
+                    C=C,
+                    d_in=d_in,
+                    d_out=d_out,
+                    bias=module.bias.data if module.bias is not None else None, # pyright: ignore[reportUnnecessaryComparison]
+                )
```
oli-clive-griffin, line 161: only pass in the tensor, not the param. I don't **think** this was causing issues, but good to be sure

spd/run_spd.py
```diff
@@ -71,11 +71,6 @@ def optimize(
         gate_hidden_dims=config.gate_hidden_dims,
         pretrained_model_output_attr=config.pretrained_model_output_attr,
     )
-
-    for param in target_model.parameters():
-        param.requires_grad = False
-    logger.info("Target model parameters frozen.")
-
```
oli-clive-griffin, line 78: this is is important change. Instead of freezing the whole model (which now contains the components via `ComponentsOrModule`), we freeze before adding these, in `ComponentModel.__init__`

spd/models/component_model.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         pretrained_model_output_attr: str | None,
     ):
         super().__init__()
+        target_model.requires_grad_(False)
```
oli-clive-griffin, line 53: Here, we freeze the target **before** adding the `ComponentsOrModule`s

PR #93, opened 2025-07-28
Add support for decomposing GPT-2 style models

casperlchristensen:
## Description

GPT-2 (and other models) are for legacy reasons implemented with a Conv1D layer which essentially does the exact same thing as a Linear layer. There are small changes needed to support them, though.

## Motivation and Context

GPT-2 is a fairly well-studied model, and given that its architecture has been mapped out pretty thoroughly, it makes sense to see how SPD fares.

## How Has This Been Tested?

With tests and local runs.

Nope!

### DIFF
--- a/spd/experiments/lm/gpt2_config.yaml
+++ b/spd/experiments/lm/gpt2_config.yaml
@@ -54,7 +54,7 @@ metrics_fns:
 
 
 # --- Pretrained model info ---
-pretrained_model_class: transformers.AutoModelForCausalLM
+pretrained_model_class: transformers.GPT2LMHeadModel
 pretrained_model_name_hf: openai-community/gpt2
 pretrained_model_output_attr: logits
 tokenizer_name: openai-community/gpt2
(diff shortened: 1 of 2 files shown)


mivanit:
just a thought -- wouldn't it make sense to try to adapt the interface to support TransformerLens models, since they've already done the work of porting a bunch of open weights models to a common architecture (including for gpt2). It may be that doing this is way too hard, I'm not actually sure

danbraunai (approved):
Nice. Definitely want to add things like this so we can support more models.

I'm scared that people are going to think this is a [torch.nn.Conv1d](https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv1d.html) layer, which is completely different with weights (in_channels, out_channels, kernel_size):
```
>>> import torch.nn as nn
>>> m = nn.Conv1d(16, 33, 1)
>>> m.weight.shape
torch.Size([33, 16, 1])
```

Perhaps it's best to alias Conv1D everywhere it's imported. Maybe RadfordConv1D or TransposedLinear?

Please add a happy path test for the new gpt2_config.yaml.

If those two things are addressed without issue, feel free to merge yourself. If you disagree or other things happen in this PR, ping me.

PR #81, opened 2025-07-25
Fix model loading

mivanit:
## Description
model loading (of freshly trained models) is broken. we should fix it, and also add tests so this doesn't happen on ~~main~~ dev in the future.

## How Has This Been Tested?
still WIP

## Does this PR introduce a breaking change?
might potentially have to change some interfaces to fix the model loading issue, because a broken change was pushed to ~~main~~ dev

PR #95, opened 2025-07-30
resid_mlp renaming and simplify naming for loading/saving

danbraunai:
## Description

- Renames ResidualMLP to ResidMLP in all class names
- Renames the task_name from residual_mlp to resid_mlp
- Uses the task_name (loaded from the experiments `TaskConfig`) when loading and saving models
- Trains new resid_mlp models. See runs below

Target model training:
- resid_mlp1: https://wandb.ai/goodfire/spd/runs/pziyck78. loss 0.000696
- resid_mlp2: https://wandb.ai/goodfire/spd/runs/any9ekl9. loss 0.000690
- resid_mlp3: https://wandb.ai/goodfire/spd/runs/6hk3uciu. loss 0.000658

SPD runs with seeds 0-9:
- resid_mlp1:
  - sweep: https://wandb.ai/goodfire/spd?nw=nk5w647mspi
    - All solve it
  - seed 0: https://wandb.ai/goodfire/spd/runs/ouia3ciq
- resid_mlp2:
  - sweep: https://wandb.ai/goodfire/spd?nw=ei1zpibz4f2
    - Most of these just solve it. A few of them have clear errors. All of them have some very small noise in the second layer.
  - seed 0: https://wandb.ai/goodfire/spd/runs/htch0mwf
- resid_mlp3:
  - sweep: https://wandb.ai/goodfire/spd?nw=nkwb8vkm15z
    - All of them are a few off. Which isn't unusual.
  - seed 0: https://wandb.ai/goodfire/spd/runs/4tb0shtw

## Related Issue
N/A

## Motivation and Context
- We use "residual_mlp" as the task name, but "resid_mlp" for variables everywhere else
- It'd be nice if we could use the standard "task_name" of resid_mlp for everything. It'd also be nice if the task name is a prefix of the experiment name, but we don't enforce that anywhere yet.

## How Has This Been Tested?
None, apart from the runs above.

Yes. Old resid_mlp models won't be loadable. We've also trained new toy models for resid_mlp so expect (hopefully minor) difference in SPD solve frequencies.


mivanit:
would it be a good idea to add the canonical run loading tests from #81 into this PR?

PR #102, opened 2025-08-04
Support data parallelism

### LINKED ISSUE #69: Add Data Parallelism support
Currently, when running SPD non-trivially size models, runs take a while and run out of memory fairly easily. On SimpleStories, for example, Lucius is currently having to run with a batchsize in the range of 50-64, which is very small. 

We think it would speed up training (and our ability to use compute productively) to be able to distribute a batch over multiple GPUs, as in data-parallelism. 

This is something of a bottleneck for key experiments, so is Important and Urgent.

danbraunai:
## Description
- Allows for data parellelism (dp) for lm experiments.
- Uses `torch.nn.parallel.DistributedDataParallel` to wrap the model and a `torch.utils.data.DistributedSampler` to wrap the dataset. This will give each rank its own data, and automatically reduce them when running .backward() in training.
- Can run with e.g. `spd-run --experiments ss_mlp --dp=8` which will run the one experiment on a node with 8 gpus and split the data evenly across all of them.
- A very rough test shows that 8 gpus is ~5 times faster than 1 gpu for ss_mlp with bs=64.
- Puts ComponentModel.forward_with_pre_forward_cache_hooks() and ComponentModel.forward_with_components() inside ComponentModel.forward() with a new `type: Literal["target", "components", "pre_forward_cache"]` TODO: change to "original" instead of "default".
- Cannot maintain determinism when producing random masks for the stochastic losses without serious costs, so we're giving up on determinism for now. This is because pytorch's RNG is not counter-based on cuda, so e.g.:
```
    device = "cpu"
    generator = torch.Generator(device=device).manual_seed(0)
    x1 = torch.rand((2,), device=device, generator=generator)
    x2 = torch.rand((2,), device=device, generator=generator)
    # cuda -> x1: [0.54, 0.12], x2: [0.98, 0.43]
    # cpu -> x1: [0.12, 0.54], x2: [0.43, 0.98]

    generator = torch.Generator(device=device).manual_seed(0)
    y = torch.rand((4,), device=device, generator=generator)
    # cuda -> y: [0.54, 0.12, 0.09, 0.87]
    # cpu -> y: [0.12, 0.54, 0.43, 0.98]
```
- We assert that each rank has the same batch size for simplicity.

## Related Issue
Closes #69 

## Motivation and Context
More speed

## How Has This Been Tested?
- Added `tests/test_distributed.py` which contains some utility tests as well as an end-to-end test which does mpi runs with dp=1 and dp=2 for several steps and shows that, when there are no stochastic losses (which are nondetermistic due to random mask generation), then we get deterministic train and eval logs as well as parameters.
- note: For some reason the ce_unrecovered metrics differ a far bit between runs in the above tests. I assume the operation used to calculate that is very prone to floating point errors. But this might be an indication to not read too much into very small differences in those metrics in general.
- Added test that alivecomponentstracker works with DDP (mocking ReduceOp.MAX and running on CPU for simplicity)
- Have also done various manual tests showing similar things.

Run report of toy models [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250808_143453--VmlldzoxMzkxMzI1MQ==).

No


spd/run_spd.py
```diff
@@ -141,23 +171,26 @@ def optimize(
...
 
         microbatch_log_data = defaultdict[str, float](float)
-        for _ in range(config.gradient_accumulation_steps):
+        for _ in range(gradient_accumulation_steps):
             batch = extract_batch_data(next(train_iterator)).to(device)
 
-            target_out, pre_weight_acts = model.forward_with_pre_forward_cache_hooks(
-                batch, module_names=model.target_module_paths
+            # Use component_model for special forward pass (DDP wrapper doesn't have this method)
+            target_out, pre_weight_acts = component_model.forward_with_pre_forward_cache_hooks(
```
danbraunai, line 178: TODO: This doesn't work with DistributedDataParallel, I think we're going to need to use `__call__` for everything and have an argument in ComponentModel.forward() indicating what time of forward pass to run (pre_cache, components, target, etc.).

danbraunai: Done.

PR #109, opened 2025-08-14
GPT-2 work from other PR

casperlchristensen:
## Description

GPT-2 (and other models) are for legacy reasons implemented with a Conv1D layer which essentially does the exact same thing as a Linear layer. There are small changes needed to support them, though.

## Motivation and Context

GPT-2 is a fairly well-studied model, and given that its architecture has been mapped out pretty thoroughly, it makes sense to see how SPD fares.

## How Has This Been Tested?

With tests and local runs.

Nope!


danbraunai (approved):
As discussed in #93, all good!

mivanit:
the `ExperimentConfig`s in the registry for `gpt2`, `ss_llama`, and `ss_gpt2` are missing a canonical run, and this causes my code to crash -- not sure why this wasn't caught by the tests.

danbraunai:
> the ExperimentConfigs in the registry for gpt2, ss_llama, and ss_gpt2 are missing a canonical run, and this causes my code to crash

Curious as to where you get this error, I can't reproduce. Though I notice that if you try and use the UVPlots, IdentityCIError, or PermutedCIPlots evals with a language model run then it will fail. I added asserts to make these fails clearer to the user in f97544a

PR #112, opened 2025-08-14
Shuffle data after each epoch

### LINKED ISSUE #104: Shuffle data each epoch
Currently I believe that we get the same data order when going through multiple epochs. It would be good if we would instead shuffle the data order each epoch.

Note that for DDP (merged in #102), you may need to use DistributedSampler.set_epoch() when starting a new epoch.

We should test that this works for dp=1 and dp>1.

danbraunai:
## Description
- Change loop_dataloader and create_data_loader so that, by default, data will be shuffled after each epoch.
- Handles both dp > 1 and dp == 1. For dp > 1, applies `sampler.set_epoch(epoch)` at the start of each epoch
- Does not affect TMS, resid_mlp, or ih, as those datasets are generated on the fly and thus have no such thing as an epoch

## Related Issue
Closes #104 

## Motivation and Context
We want more variety in the dataset after each epoch. We often use multiple epochs in simplestories.

## How Has This Been Tested?
Manually tested that we indeed get new data each epoch with dp==1 and dp>1. It's very cumbersome to write a proper test for this with mpi for dp>1, so not bothering. I don't think it's important enough.

Had to update the existing test_distributed_determinicity to turn off shuffling, because with the shuffling on the non-distributed version gives different batch data to the distributed version.

Yes. Runs will produce different results given that the datasets will now shuffle after each epoch.

### DIFF
--- a/spd/data.py
+++ b/spd/data.py
@@ -236,8 +236,6 @@ def loop_dataloader[T](dl: DataLoader[T]):
     Ensures that each epoch gets different data, even when using a distributed sampler.
     """
     epoch = 0
-    if isinstance(dl.sampler, DistributedSampler):
-        dl.sampler.set_epoch(epoch)
     dl_iter = iter(dl)
     while True:
         try:


danbraunai:
@claude what do you think about this PR? Interested if there are any potential errors you see with it, or ways you think it could be managed better. Feel free to be a ruthless as you'd like, hard truths will make this the most effective PR review it could be.

Issue #111, opened 2025-08-14
Handle streaming datasets with dp > 1


danbraunai:
Currently, we don't support streaming with dp > 1. We should do this. I'm unsure of how to do this with the DistributedSampler. It's possible we have to make our own sampler. E.g. like https://github.com/karpathy/llm.c/blob/master/train_gpt2.py

We need this if we want to decompose gpt2 with e.g. openwebtext.

danbraunai:
Closed in #113

PR #113, opened 2025-08-15
Support streaming and tokenized datasets with and without ddp

danbraunai:
## Description
Support streaming datasets with and without ddp, as well as tokenized datasets.

- When dp>1 and streaming=True, we shard the dataset across ranks and use the default sampler.
    If the dataset has fewer dataset shards than we have ddp ranks, we split up the dataset
    by example. We also use dataset.set_epoch(epoch) during training.
- When dp>1 and streaming=False, we use a DistributedSampler and run
    sampler.set_epoch(epoch) during training.
- When dp=1 and streaming=True, we use the default sampler and run
    dataset.set_epoch(epoch) during training.
- When dp=1 and streaming=False, we use the default sampler and set shuffle=True on the
   DataLoader.

Note: When I run with streaming, it's quite slow. It goes fast for ~18 steps, then a slow step, then back to fast. This is likely due to the dataset.map function which does the tokenization. I'm sure that better settings could speed this up, but I think streaming will be sufficiently niche as to not warrant it.

## Related Issue
#111 

## Motivation and Context
We want people to be able to decompose very large datasets.

## How Has This Been Tested?
Testing this would be very time consuming and involve refactors. Right now, I don't think streaming should be a common use case, given that it's also quite slow. People should just set it to false and have the dataset cached.

For this reason, I didn't write tests and just gave warnings to the user that the streaming=True is not well tested and that it's recommended to not use it.

No

### DIFF
--- a/spd/data.py
+++ b/spd/data.py
@@ -251,8 +251,7 @@ def create_data_loader(
 
     # Ensure determinicity when not distributed and not streaming
     generator = torch.Generator(device="cpu")
-    if not dataset_config.streaming and not is_ddp:
-        generator.manual_seed(seed)
+    generator.manual_seed(seed)
 
     loader = DataLoader[Dataset | IterableDataset](
         torch_dataset,  # pyright: ignore[reportArgumentType]
--- a/spd/data.py
+++ b/spd/data.py
@@ -194,7 +194,12 @@ def create_data_loader(
 
     if dataset_config.streaming:
         assert isinstance(dataset, IterableDataset)
+        logger.warning(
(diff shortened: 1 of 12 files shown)


danbraunai:
@claude what do you think about this PR? Interested if there are any potential errors you see with it, or ways you think it could be managed better. Feel free to be a ruthless as you'd like, hard truths will make this the most effective PR review it could be.

danbraunai:
@claude I've made several changes since you last looked. Could you review these changes and the PR in general. I'm not concerned about potential nondeterminicity issues when streaming=True

PR #117, opened 2025-08-18
Update checks.yaml

aravindan888:
Fixes#116


aravindan888:
changed the version to 5.0.8, can you check if it is working now

danbraunai:
I think you need to remove the Ubuntu suffix

danbraunai:
Seems to fail for an unknown reason. You might be better off debugging this in a small private repo.

PR #140, opened 2025-09-08
remove init_from_target_weight; init U,V in init

Laplace418:
## Description
Removes deprecated init_from_target_weight and initializes V and U in Components.__init__.
Removes all call sites and updates tests.
Branch was reset to origin/main; only minimal changes included.
Files: spd/models/components.py, spd/models/component_model.py, tests/test_component_model.py.

## Motivation and Context
init_from_target_weight doesn't scale right when the number of components gets large, leading to problems on LLMs

## How Has This Been Tested?
Ran the code on the seven toy models (tms 5-2, tms 5-2 identity, tms 40-10, tms 40-10 identity, resid mlp1, resid mlp2, resid mlp3). All perfect except resid mlp3, which had 1 flaw. resid mlp3 has flaws on main as well though, so I think that's ok. Report: https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250908_114347--VmlldzoxNDMwMDE5NQ==

Runs will behave differently because the initalisation is different now.

### DIFF
--- a/spd/models/component_model.py
+++ b/spd/models/component_model.py
@@ -234,14 +234,12 @@ class ComponentModel(LoadableModule):
                     d_out=d_out,
                     bias=module.bias.data if module.bias is not None else None,  # pyright: ignore[reportUnnecessaryComparison]
                 )
-                # Components are initialized in __init__
             elif isinstance(module, nn.Embedding):
                 component = EmbeddingComponents(
                     C=C,
                     vocab_size=module.num_embeddings,
                     embedding_dim=module.embedding_dim,
                 )
-                # Components are initialized in __init__
             elif isinstance(module, RadfordConv1D):
                 d_in, d_out = module.weight.shape
                 component = LinearComponents(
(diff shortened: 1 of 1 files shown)


spd/models/component_model.py
```diff
@@ -234,14 +234,14 @@ def _patch_modules(
                     d_out=d_out,
                     bias=module.bias.data if module.bias is not None else None,  # pyright: ignore[reportUnnecessaryComparison]
                 )
-                component.init_from_target_weight(module.weight.T)
+                # Components are initialized in __init__
```
danbraunai, line 237: Remove comments like this

spd/models/components.py
```diff
@@ -92,41 +92,14 @@ def __init__(self, C: int, v_dim: int, u_dim: int):
         self.C = C
         self.V = nn.Parameter(torch.empty(v_dim, C))
         self.U = nn.Parameter(torch.empty(C, u_dim))
+        init_param_(self.V, fan_val=v_dim, nonlinearity="linear")
+        init_param_(self.U, fan_val=C, nonlinearity="linear")
```
danbraunai, line 96: Just to confirm, the fan_val corresponds to the input dimension? Data goes through V first. Resolve comment if agree.

danbraunai (approved):
All good. A few comments. Can merge after you've resolved them.

PR #141, opened 2025-09-08
Allow inserting identity matrices before any module

### LINKED ISSUE #131: Add ability to insert identity matrices
We know that the ideal decomposition for an identity matrix in the target model is to have one component per dimension. We can use this fact by inserting identity matrices at various points in the target model when doing an SPD run, and verifying that indeed we do get one component per dimension for the identity matrices.

It's worth thinking very carefully about how to implement this, as I imagine there is large variance in how clean different implementations might be.

danbraunai:
## Description
- `identity_module_patterns` is added to the main config, indicating the modules in which we want to insert an identity transformation just before it is called.
- We modify `ComponentsOrModule` by adding an optional `identity_components` attribute, and making the existing `components` attribute optional. Now, we modify a module in the original target model if that module exists in either `target_module_patterns` or `identity_module_patterns`. If it exists in both, we just have the single `ComponentsOrModule` object with non-None values for `components` and `identity_components`. Then, the forward pass of `ComponentsOrModule` does:
  - If forward_mode == "original",   just pass through the original model as before.
  - if forward_mode == "components", then if identity_components is set, pass through an identity component instance (which is just a LinearComponent with no bias) and then either pass through the original model if this was _just_ an identity transformation (`ComponentsOrModule.components is None`), or pass through the components if this was an identity + components transformation (`ComponentsOrModule.components is not None`). 

- Add CIMeanPerComponent eval. For identity matrix insertion, there should be a sharp dropoff on the mean ci value for components larger than the rank. An example below from this [run](https://wandb.ai/goodfire/spd?nw=lz6d9czllua) (although note the dropoff isn't extremely sharp, which isn't surprising given that this run doesn't converge to a perfect CI solution
image

## Related Issue
Closes #131.

## Motivation and Context
If we "pretend" that the original matrix had an identity layer, we think that we should expect SPD to learn a decomposition of it where, for an identity transformation of size `d`, we get exactly `d` components that are ever active and that perform the transformation.

## How Has This Been Tested?
Added various unittests in `tests/test_component_model.py` for behaviour related to our identity and regular components.

- [Run](https://wandb.ai/goodfire/spd/runs/85d2bk11) showing that tms_40-10 with an added identity_module_pattenrs=["linear2"] gets extremely similar results to [tms_40-10-id](https://wandb.ai/goodfire/spd/runs/dsn94977), and the identity decomposition looks as expected.
- [Run](https://wandb.ai/goodfire/spd/runs/pogwh7ep) showing that resid_mlp1 with an added identity_module_patterns=["layers.0.mlp_out"] is on its way to converging to a dense CI matrix but doesn't quite get there. 

No

### DIFF
--- a/spd/losses.py
+++ b/spd/losses.py
@@ -170,12 +170,12 @@ def calc_masked_recon_loss(
     return loss
 
 
-def _calc_tensors_mse(
+def _calc_tensors_squared_error(
     params1: dict[str, Float[Tensor, "d_in d_out"]],
     params2: dict[str, Float[Tensor, "d_in d_out"]],
     device: str,
 ) -> Float[Tensor, ""]:
-    """Calculate the MSE between params1 and params2, summing over the d_in and d_out dimensions.
+    """Calculate the squared error between params1 and params2, summing over d_in and d_out dims.
 
     Args:
         params1: The first set of parameters
@@ -209,7 +209,7 @@ def calc_faithfulness_loss(
             target_params[id_name] = torch.eye(id_mat.shape[0], device=device, dtype=id_mat.dtype)
             assert component_params[id_name].shape == target_params[id_name].shape
 
-    faithfulness_loss = _calc_tensors_mse(
+    faithfulness_loss = _calc_tensors_squared_error(
         params1=component_params,
         params2=target_params,
         device=device,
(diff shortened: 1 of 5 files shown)


spd/losses.py
```diff
@@ -138,12 +138,10 @@ def calc_masked_recon_layerwise_loss(
...
     for mask_info in masks:
-        for component_name in model.components:
-            modified_out = model(
-                batch,
-                mode="components",
-                masks={component_name: mask_info[component_name]},
-            )
+        for comp_name, mask in mask_info.items():
+            # TODO: Write a test showing that passing a mask for a ComponentOrModule which has
+            # both components and identity components works as expected.
```
danbraunai, line 143: Flagging this. Should be done before merging. Should be some other tests of this too.

PR #146, opened 2025-09-10
feat: Add SubsetReconstructionLoss evaluation metric

nathu-goodfire:
## Description
  Adds evaluation logging to compute reconstruction loss where we only use SPD components on a subset of the layers. We support this via regex matching.

  ## Related Issue
  Related to issues in current SPD runs on SimpleStories where the final MLP down layer has high L0.

  ## Motivation and Context
  We noticed the final MLP down layer has high L0 in SimpleStories runs. These runs fail sanity checks where using SPD components on earlier layers but the original model weights on the high l0 mlp down module leads to worse reconstruction loss. By computing reconstruction loss on subsets, we can catch this issue earlier in training.

  ## How Has This Been Tested?
  Added a new config reproducing Lucius's current best SimpleStories Llama SPD run with these metrics. Run looks good and metrics already show higher recon loss when not using SPD components to replace the final MLP.
  https://wandb.ai/goodfire/spd/reports/test-subset-metrics--VmlldzoxNDMyNTA2OA

  No

### DIFF
--- a/spd/eval.py
+++ b/spd/eval.py
@@ -7,6 +7,7 @@ These can be selected and configured in the Config.
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Iterator, Mapping
+from fnmatch import fnmatch
 from typing import Any, ClassVar, override
 
 import einops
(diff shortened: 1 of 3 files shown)


danbraunai (approved):
Few changes suggested. Can merge if addressed.

The [report](https://wandb.ai/goodfire/spd/reports/test-subset-metrics--VmlldzoxNDMyNTA2OA) you linked doesn't seem to show the new eval metrics you added.

I'm not loving how many line plots this adds to wandb. Similarly, not loving the existing l0 line plots for every layer. Perhaps we just want a bar chart showing the metric for all layers that just changes each step? This may be best done in wandb reports/workspaces rather than in code. Although it is nice to have the same images saved locally that we use on wandb. But it might be much slower if we create these combined bar charts and then upload them to wandb.

danbraunai:
It looks like this run using continuous sampling and not binomial sampling?

danbraunai:
Lucius just mentioned also that he would love a metric that is just all the L0 values added together.

PR #148, opened 2025-09-11
Layerwise global ci function

danbraunai:
## Description
- Adds `gate_type: "layerwise_global_mlp"` as a config option. This is not placed in any of the default configs given that it's not clearly better.
- The gate_type above corresponds to `spd.models.components.LayerwiseGlobalGateMLP`. The basic idea is that it's a pure MLP which takes as input the input to the module it is decomposing, and outputs a vector of size C. Unlike VectorGateMLPs and GateMLPs, we don't split it up into separate networks for each subcomponent C.

Based on the runs we did, this wasn't very obviously better than vector-mlp, although it was comparable. Further investigation is needed to see if variations on this, such as multiple hidden layers, performs better than vector-mlp (with the same variations applied).

Setting a gate_hidden_dims value of around 3000 or 1000 got reasonable results for ss_llama.

## Related Issue
Relates to #145 

## Motivation and Context
A layerwise global gate is simpler than VectorMLPs and GateMLPs, which use an individual network for each subcomponent. We thought that it might also be much faster, but the analysis shown doesn't indicate this.

## How Has This Been Tested?
- Language model [report](https://wandb.ai/goodfire/spd/reports/Global-layerwise-CI-function-comparisons--VmlldzoxNDQyNzI1OQ)
  - Summary: gate3k with importance_min=2e-4 and 3e-4 are pretty good. Comparable to the vector-mlp with 12 hidden dims (also a similar number of total weights in the gate). Full run log at end of description
- Toy models
  - Summary: it works for specific combinations of hyperparams for all models except for resid_mlp3. On resid_mlp3, the results are very bad. Possible better hyperparams will fix it. Full run log at end of description

## Does this PR introduce a breaking change?
No

### ss_llama
- [[Sweep](https://wandb.ai/goodfire/spd?nw=vchdm4bg2lb)](https://wandb.ai/goodfire/spd?nw=vchdm4bg2lb). global_layerwise gate 1000 dims over imp_min=[5e-4, 3e-4, 2e-4, 1e-4]
    - None of these are strictly better than vector-mlp.
        - 2e-4 run is the closest in terms of CE losses, but has worse L0
        - Otherwise, you trade off L0 for CE losses.
- [[Sweep](https://wandb.ai/goodfire/spd?nw=6pyqgyhc2eo)](https://wandb.ai/goodfire/spd?nw=6pyqgyhc2eo). global_layerwise gate 3000 dims over imp_min=[5e-4, 3e-4, 2e-4, 1e-4]

### Toy models
[...]

PR #151, opened 2025-09-12
Remove faithfulness loss with a delta component

### LINKED ISSUE #24: Faithfulness loss threshold for spd-evals is on the wrong scale and far too permissive
**Migrated from:** goodfire-ai/spd-gf#37
**Original author:** @Laplace418

---

Faithfulness loss for spd-evals is set to 1e-03.

- This doesn't generally mean anything, because faithfulness loss is on a completely different scale for different models
- For many of the toy models we have, this is waaay too permissive.

We need a different scale and a better default threshold. Options:

1. Take the square root of the faithfulness loss and divide it by the L2 weight norm of the original model weights. Might also need to multiply by the number of parameters before square rooting, I think we're currently normalising with that. This way, we get to see the size of the divergence relative to the size of the original model
2. Use a threshold based on the unmasked reconstruction loss or loss recovered instead.

For finding good thresholds for each toy model, you should make sure that

1. Randomly initialised components score above the threshold
2. Our current canonical solutions to each toy model score below the threshold

### LINKED ISSUE #132: Get rid of faithfulness loss
We may be able to get rid of the faithfulness loss entirely by having an extra full-rank component that spans all layers and that is always on.

It's worth thinking very carefully about how to implement this, as I imagine there is large variance in how clean different implementations might be. Speak to Lucius as he has some ideas.

danbraunai:
## Description
- Adds use_delta_component config argument. This controls whether, for each layer, we use an extra component representing the difference between the sum of all the "regular" components and the target model. **Defaults to true**.
- We now calculate these weight deltas in the training loop and then pass them to a new faithfulness loss calculation as well as the component forward pass.
- calc_stochastic_masks now returns a new mask that is to be applied to this special delta component.
- In the component forward pass, we now add to the old forward pass the output of this new masked delta component.
- Raise error if user tries to decompose embedding layers with `gate_type != mlp`. More work would need to be done to handle different gates with embedding layers that take integer inputs.

NOTE: This breaks some toy models (e.g. tms_5-2) with the current configs. See report [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250917_082928--VmlldzoxNDQyMDUzNg==). Hyperparameter optimization on toy models will be bundled in with #148.

## Related Issue
Closes #132; closes #24

## Motivation and Context
One less loss function. What's not to like?

## How Has This Been Tested?
- Updated test_spd_losses to handle the new structure for faithfulness loss and added tests for calc_weight_deltas.
- No new tests added to show that components handle weight_deltas properly. This is not ideal. Feel free to suggest tests if reviewing this. Not immediately obvious to me what tests would be useful here.

Runs showing that it's the same as Lucius' branch
- [Run](https://wandb.ai/goodfire/spd/runs/1aqil6mu) from this branch. NOTE: The SubsetReconstruction losses in this run don't use the weight delta masks (later commits added this).
- [Run](https://wandb.ai/goodfire/spd/runs/ec1eydv2) from Lucius' branch

- [Report](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250915_144742--VmlldzoxNDM5NDMzOA==) for run with all toy models.
  - resid_mlp3 doesn't work, although it's always been a little dodgy. Can probably get it working or near working with more sweeps. p-annealing might help it.
  - Ignore the tms_40-10-id run and consider [this run](https://wandb.ai/goodfire/spd/runs/gmxxu1g9) instead.

### Testing runtime
There exists a different implementation of this idea which doesn't require materialising all of the weight deltas. However, that implementation is more complex and harder to understand. We've gone with the simpler but theoretically more expensive implementation here. We thus benchmarked the runtime of the implementation in this PR, compared with an implementation which doesn't calculate the weight deltas at all (and removes the call to weight deltas everywhere).
- Runtime of ss_llama with bs=8 (dp=1) in this PR: 7h30m
- Runtime of ss_llama with bs=8 (dp=1) when removing weight delta calculation and use: 7h15m

So, at best, the fancier implementation would not meaningfully save in runtime. It's possible that a heavily optimised future version may want to use the fancier implementation for runtime and memory reasons, but we'll leave that for the future.

An extremely rough version of the equations involved in producing the fancier implementation is below:
$$$
W-sum_i U_i V_i = Delta
L_faith=Delta^2

sum_i m(x)_i U_i V_i +r(x) Delta
r(x) \in (0,1)
sum_i m(x)_i U_i V_i +r(x) (W-sum_i U_i V_i )
sum_i U_j,i m(x)_i V_i,k +r(x) Delta_j,k
 
sum_i U_j,i m(x)_i V_i,k +r(x) (W-sum_i U_j,i V_i,k )
sum_i U_j,i (m(x)_i-r(x)) V_i,k +r(x) W
m’(x)_i=(m(x)_i-r(x))
sum_i U_j,i m’(x)_i V_i,k +r(x) W
 
out_spd +r(x)out_target
out_target=Wx+b
out_spd= U (m-r)Vx+b

out_spd +r(x)out_target=U (m-r)Vx+r(x) Wx+(1+r(x))b
$$$

Yes. By default we have `use_delta_component: True`. Also note that all config files now specify "continuous" rather than "binomial" sampling as this seems to be on par with binomial for LM experiments after this PR.

### DIFF
--- a/spd/eval.py
+++ b/spd/eval.py
@@ -635,11 +635,11 @@ class SubsetReconstructionLoss(StreamingEval):
             active = [m for m in all_modules if any(fnmatch(m, p) for p in patterns)]
 
             outputs = self._get_masked_model_outputs(
-                batch,
-                masks_list,
-                weight_deltas,
-                active,
-                all_modules,
+                batch=batch,
+                masks_list=masks_list,
+                weight_deltas=weight_deltas,
+                active=active,
+                all_modules=all_modules,
             )
             kl_losses = [kl_vs_target(out) for out in outputs]
             ce_losses = [ce_vs_labels(out) for out in outputs]
@@ -658,11 +658,11 @@ class SubsetReconstructionLoss(StreamingEval):
             active = [m for m in all_modules if not any(fnmatch(m, p) for p in exclude_patterns)]
 
             outputs = self._get_masked_model_outputs(
-                batch,
-                masks_list,
-                weight_deltas,
-                active,
-                all_modules,
+                batch=batch,
+                masks_list=masks_list,
+                weight_deltas=weight_deltas,
+                active=active,
+                all_modules=all_modules,
             )
             kl_losses = [kl_vs_target(out) for out in outputs]
             ce_losses = [ce_vs_labels(out) for out in outputs]
--- a/spd/losses.py
+++ b/spd/losses.py
@@ -7,7 +7,7 @@ from jaxtyping import Float, Int
 from torch import Tensor
 
 from spd.configs import Config
-from spd.mask_info import ComponentsMaskInfo, make_mask_infos
+from spd.mask_info import ComponentsMaskInfo, WeightDeltaAndMask, make_mask_infos
 from spd.models.component_model import ComponentModel
 from spd.models.components import Components, ComponentsOrModule, EmbeddingComponents
 from spd.utils.component_utils import calc_stochastic_masks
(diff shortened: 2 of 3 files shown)


spd/run_spd.py
```diff
@@ -171,6 +171,9 @@ def optimize(
 
         microbatch_log_data: defaultdict[str, float] = defaultdict(float)
         current_p = config.pnorm  # Initialize with default value
+
+        weight_deltas = calc_weight_deltas(component_model, device)
```
nathu-goodfire, line 175: As is this does not work with grad accumlation:
either need to make a new weight delta for each accumlation step. 
or add a retain_graph=True somewhere.

danbraunai:
@claude could you review this PR? I'm interested in high level questions about the structure of the new features and any ways that it could be improved. Including object naming. I want critical comments, do feel comfortable to give me the hard truths.

spd/run_spd.py
danbraunai: Ohhh I see haha. Fixed (just moved it inside the grad accum loop).

oli-clive-griffin:
Continuing this. Additions:

- unify `weight_delta` and `weight_delta_and_mask` into `weight_delta_and_mask`
- simplified `SubsetReconstructionLoss` repeated code around that masking logic
- generally naming tidy ups to make the relationship between identity, component etc. more clear

oli-clive-griffin (approved):
Looks good, can't really think of great ways to make it a lot nicer, think we're good for now re effective:hacky ratio

PR #154, opened 2025-09-17
Geometric similarity comparison between two trained models

leesharkey:
## Description
This PR introduces a standalone script for post-hoc geometric similarity comparison between the learned components of two different SPD model runs.

The primary motivation is to verify if SPD consistently discovers similar "ground truth" solutions across runs with different random seeds or other variations.

Key Features:
- A new script, spd/scripts/compare_models/compare_models.py, allows for comparing models after training is complete.
- Computes the mean max absolute cosine similarity between the rank-one component matrices of the two models. similarity metrics (mean, std, min, max) per layer and across all layers get saved in a folder that is (by default) located in the same directory as the script itself.
- Activation density filtering to ensure that only sufficiently active components are included in the comparison. 
- WandB & Local Model Support: Models can be loaded directly from WandB run paths or local directories.
- A new README.md is included in the script's subdirectory with detailed usage instructions.

## Related Issue
None

## Motivation and Context
We would like this in order to be able to verify if SPD yields similar solutions using different random seeds (or other changes), which it should if it's finding some 'ground truth' canonical solution. 

## How Has This Been Tested?
This has been tested by running the new evaluations on a representative sample of toy models and LMs:

- TMS 5-2 with identity
- ss_llama on a single GPU. 

I explored different 'density thresholds'. A low nonzero value is appropriate (e.g. 0.001). If the value is too high such that no subcomponents activate densely enough, then the metric cannot be calculated.

Configs and reference models are provided in spd/scripts/compare_models/compare_models_config.yaml. 

### Comparing two different TMS 5-2-id models:

```spd/scripts/compare_models/compare_models_config.yaml
current_model_path: "wandb:goodfire/spd/runs/667z2n1b"
reference_model_path: "wandb:goodfire/spd/runs/vh4yszsd"
```

Similarity metrics:
  mean_max_abs_cosine_sim/linear1: 0.9995
  max_abs_cosine_sim_std/linear1: 0.0002
  max_abs_cosine_sim_min/linear1: 0.9992
  max_abs_cosine_sim_max/linear1: 0.9998
  mean_max_abs_cosine_sim/linear2: 0.9995
  max_abs_cosine_sim_std/linear2: 0.0002
  max_abs_cosine_sim_min/linear2: 0.9992
  max_abs_cosine_sim_max/linear2: 0.9998
  mean_max_abs_cosine_sim/hidden_layers.0: 0.9210
  max_abs_cosine_sim_std/hidden_layers.0: 0.0137
  max_abs_cosine_sim_min/hidden_layers.0: 0.9113
  max_abs_cosine_sim_max/hidden_layers.0: 0.9306
  mean_max_abs_cosine_sim/all_layers: 0.9733
  max_abs_cosine_sim_std/all_layers: 0.0047
  max_abs_cosine_sim_min/all_layers: 0.9699
  max_abs_cosine_sim_max/all_layers: 0.9767

### Comparing a SimpleStories model with itself

Using
```spd/scripts/compare_models/compare_models_config.yaml
current_model_path: "wandb:goodfire/spd/runs/4r8yn2zt"
reference_model_path: "wandb:goodfire/spd/runs/4r8yn2zt"
```

Similarity metrics:
  mean_max_abs_cosine_sim/model.layers.0.self_attn.q_proj: 1.0000
  max_abs_cosine_sim_std/model.layers.0.self_attn.q_proj: 0.0000
  max_abs_cosine_sim_min/model.layers.0.self_attn.q_proj: 1.0000
  max_abs_cosine_sim_max/model.layers.0.self_attn.q_proj: 1.0000
  mean_max_abs_cosine_sim/model.layers.0.self_attn.k_proj: 1.0000
  max_abs_cosine_sim_std/model.layers.0.self_attn.k_proj: 0.0000
  max_abs_cosine_sim_min/model.layers.0.self_attn.k_proj: 1.0000
  max_abs_cosine_sim_max/model.layers.0.self_attn.k_proj: 1.0000
  mean_max_abs_cosine_sim/model.layers.0.self_attn.v_proj: 1.0000
  max_abs_cosine_sim_std/model.layers.0.self_attn.v_proj: 0.0000
  ...
  mean_max_abs_cosine_sim/all_layers: 1.0000
  max_abs_cosine_sim_std/all_layers: 0.0000
  max_abs_cosine_sim_min/all_layers: 1.0000
  max_abs_cosine_sim_max/all_layers: 1.0000

### Comparing two different, barely trained SimpleStories models

Using
```spd/scripts/compare_models/compare_models_config.yaml
current_model_path: "wandb:goodfire/spd/runs/4r8yn2zt"
reference_model_path: "wandb:goodfire/spd/runs/2lq9dpnb"
```

Similarity metrics:
  mean_max_abs_cosine_sim/model.layers.0.self_attn.q_proj: 0.0581
  max_abs_cosine_sim_std/model.layers.0.self_attn.q_proj: 0.0092
  max_abs_cosine_sim_min/model.layers.0.self_attn.q_proj: 0.0398
  max_abs_cosine_sim_max/model.layers.0.self_attn.q_proj: 0.1330
  mean_max_abs_cosine_sim/model.layers.0.self_attn.k_proj: 0.0811
  max_abs_cosine_sim_std/model.layers.0.self_attn.k_proj: 0.0126
  max_abs_cosine_sim_min/model.layers.0.self_attn.k_proj: 0.0543
  max_abs_cosine_sim_max/model.layers.0.self_attn.k_proj: 0.1709
  mean_max_abs_cosine_sim/model.layers.0.self_attn.v_proj: 0.0812
  max_abs_cosine_sim_std/model.layers.0.self_attn.v_proj: 0.0124
  max_abs_cosine_sim_min/model.layers.0.self_attn.v_proj: 0.0533
  ...
  mean_max_abs_cosine_sim/all_layers: 0.0550
  max_abs_cosine_sim_std/all_layers: 0.0085
  max_abs_cosine_sim_min/all_layers: 0.0365
  max_abs_cosine_sim_max/all_layers: 0.1085

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,17 +82,11 @@ python spd/experiments/tms/tms_decomposition.py spd/experiments/tms/tms_5-2_conf
 
 ### Model Comparison
 
-For post-hoc analysis of completed runs, use the model comparison script to compute geometric similarities between subcomponents:
+Use the model comparison script to analyse (post hoc) the geometric similarities between subcomponents of two different models:
 
 ```bash
-# Using config file
 python spd/scripts/compare_models/compare_models.py spd/scripts/compare_models/compare_models_config.yaml
-
-# Using command line arguments
-python spd/scripts/compare_models/compare_models.py --current_model_path="wandb:..." --reference_model_path="wandb:..."
 ```
-
-The comparison script supports both wandb and local model paths, and calculates mean max absolute cosine similarity metrics between learned subcomponents in different runs.
 
 See `spd/scripts/compare_models/README.md` for detailed usage instructions.
 
--- a/spd/scripts/compare_models/compare_models.py
+++ b/spd/scripts/compare_models/compare_models.py
@@ -8,7 +8,7 @@ Usage:
     python spd/scripts/compare_models/compare_models.py --current_model_path="wandb:..." --reference_model_path="wandb:..."
 """
 
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from pathlib import Path
 from typing import Any
 
@@ -37,17 +37,16 @@ class CompareModelsConfig(BaseModel):
     )
 
     density_threshold: float = Field(
-        default=0.001,
-        description="Minimum activation density for components to be included in comparison",
+        ..., description="Minimum activation density for components to be included in comparison"
     )
     n_eval_steps: int = Field(
-        default=5, description="Number of evaluation steps to compute activation densities"
+        ..., description="Number of evaluation steps to compute activation densities"
     )
 
-    eval_batch_size: int = Field(default=32, description="Batch size for evaluation data loading")
-    shuffle_data: bool = Field(default=False, description="Whether to shuffle the evaluation data")
+    eval_batch_size: int = Field(..., description="Batch size for evaluation data loading")
+    shuffle_data: bool = Field(..., description="Whether to shuffle the evaluation data")
     ci_alive_threshold: float = Field(
-        default=0.0, description="Threshold for considering components as 'alive'"
+        ..., description="Threshold for considering components as 'alive'"
     )
 
     output_dir: str | None = Field(
@@ -55,38 +54,28 @@ class CompareModelsConfig(BaseModel):
         description="Directory to save results (defaults to 'out' directory relative to script location)",
     )
 
-    device: str = Field(
-        default="auto", description="Device to run comparison on (Options: 'auto', 'cuda', 'cpu')"
-    )
-
 
 class ModelComparator:
     """Compare two SPD models for geometric similarity between subcomponents."""
 
-    def __init__(
-        self,
-        config: CompareModelsConfig,
-    ):
+    def __init__(self, config: CompareModelsConfig):
         """Initialize the model comparator.
 
         Args:
             config: CompareModelsConfig instance containing all configuration parameters
         """
         self.config = config
-        self.current_model_path = config.current_model_path
-        self.reference_model_path = config.reference_model_path
         self.density_threshold = config.density_threshold
+        self.device = get_device()
 
-        self.device = get_device() if config.device == "auto" else config.device
-
-        logger.info(f"Loading current model from: {self.current_model_path}")
+        logger.info(f"Loading current model from: {config.current_model_path}")
         self.current_model, self.current_config = self._load_model_and_config(
-            self.current_model_path
+            config.current_model_path
         )
 
-        logger.info(f"Loading reference model from: {self.reference_model_path}")
+        logger.info(f"Loading reference model from: {config.reference_model_path}")
         self.reference_model, self.reference_config = self._load_model_and_config(
-            self.reference_model_path
+            config.reference_model_path
         )
 
     def _load_model_and_config(self, model_path: str) -> tuple[ComponentModel, Config]:
@@ -99,36 +88,38 @@ class ModelComparator:
 
         return model, run_info.config
 
-    def create_eval_data_loader(self, config: Config) -> Iterator[Any]:
+    def create_eval_data_loader(self) -> Iterator[Any]:
         """Create evaluation data loader using exact same patterns as decomposition scripts."""
-        task_config = config.task_config
-        task_name = task_config.task_name
+        task_name = self.current_config.task_config.task_name
 
-        if task_name == "tms":
-            return self._create_tms_data_loader(config)
-        elif task_name == "resid_mlp":
-            return self._create_resid_mlp_data_loader(config)
-        elif task_name == "lm":
-            return self._create_lm_data_loader(config)
-        elif task_name == "induction_head":
-            return self._create_ih_data_loader(config)
-        else:
+        data_loader_fns: dict[str, Callable[[], Iterator[Any]]] = {
+            "tms": self._create_tms_data_loader,
+            "resid_mlp": self._create_resid_mlp_data_loader,
+            "lm": self._create_lm_data_loader,
+            "induction_head": self._create_ih_data_loader,
+        }
+
+        if task_name not in data_loader_fns:
             raise ValueError(
-                f"Unsupported task type: {task_name}. Supported types: tms, lm, resid_mlp, induction_head"
+                f"Unsupported task type: {task_name}. Supported types: {', '.join(data_loader_fns.keys())}"
             )
 
-    def _create_tms_data_loader(self, config: Config) -> Iterator[Any]:
+        return data_loader_fns[task_name]()
+
+    def _create_tms_data_loader(self) -> Iterator[Any]:
         """Create data loader for TMS task."""
         from spd.experiments.tms.configs import TMSTaskConfig
         from spd.experiments.tms.models import TMSTargetRunInfo
         from spd.utils.data_utils import DatasetGeneratedDataLoader, SparseFeatureDataset
 
-        assert isinstance(config.task_config, TMSTaskConfig)
-        task_config = config.task_config
+        assert isinstance(self.current_config.task_config, TMSTaskConfig)
+        task_config = self.current_config.task_config
 
-        assert config.pretrained_model_path, "pretrained_model_path must be set for TMS models"
+        assert self.current_config.pretrained_model_path, (
+            "pretrained_model_path must be set for TMS models"
+        )
 
-        target_run_info = TMSTargetRunInfo.from_path(config.pretrained_model_path)
+        target_run_info = TMSTargetRunInfo.from_path(self.current_config.pretrained_model_path)
 
         dataset = SparseFeatureDataset(
             n_features=target_run_info.config.tms_model_config.n_features,
@@ -146,19 +137,21 @@ class ModelComparator:
             )
         )
 
-    def _create_resid_mlp_data_loader(self, config: Config) -> Iterator[Any]:
+    def _create_resid_mlp_data_loader(self) -> Iterator[Any]:
         """Create data loader for ResidMLP task."""
         from spd.experiments.resid_mlp.configs import ResidMLPTaskConfig
         from spd.experiments.resid_mlp.models import ResidMLPTargetRunInfo
         from spd.experiments.resid_mlp.resid_mlp_dataset import ResidMLPDataset
         from spd.utils.data_utils import DatasetGeneratedDataLoader
 
-        assert isinstance(config.task_config, ResidMLPTaskConfig)
-        task_config = config.task_config
+        assert isinstance(self.current_config.task_config, ResidMLPTaskConfig)
+        task_config = self.current_config.task_config
 
-        assert config.pretrained_model_path, "pretrained_model_path must be set for ResidMLP models"
+        assert self.current_config.pretrained_model_path, (
+            "pretrained_model_path must be set for ResidMLP models"
+        )
 
-        target_run_info = ResidMLPTargetRunInfo.from_path(config.pretrained_model_path)
+        target_run_info = ResidMLPTargetRunInfo.from_path(self.current_config.pretrained_model_path)
 
         dataset = ResidMLPDataset(
             n_features=target_run_info.config.resid_mlp_model_config.n_features,
@@ -178,18 +171,18 @@ class ModelComparator:
             )
         )
 
-    def _create_lm_data_loader(self, config: Config) -> Iterator[Any]:
+    def _create_lm_data_loader(self) -> Iterator[Any]:
         """Create data loader for LM task."""
         from spd.data import DatasetConfig, create_data_loader
         from spd.experiments.lm.configs import LMTaskConfig
 
-        assert config.tokenizer_name, "tokenizer_name must be set"
-        assert isinstance(config.task_config, LMTaskConfig)
-        task_config = config.task_config
+        assert self.current_config.tokenizer_name, "tokenizer_name must be set"
+        assert isinstance(self.current_config.task_config, LMTaskConfig)
+        task_config = self.current_config.task_config
 
         dataset_config = DatasetConfig(
             name=task_config.dataset_name,
-            hf_tokenizer_path=config.tokenizer_name,
+            hf_tokenizer_path=self.current_config.tokenizer_name,
             split=task_config.eval_data_split,
             n_ctx=task_config.max_seq_len,
             is_tokenized=task_config.is_tokenized,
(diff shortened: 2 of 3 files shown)


spd/experiments/lm/ss_llama_single_with_comparison_config.yaml
danbraunai, line 1: I think it makes sense not to have new config files in `main` just to show people how to make comparisons. Might be nicer just adding some commented out version of the comparison eval in the main ss_llama config.

spd/experiments/lm/ss_llama_single_with_comparison_config.yaml
```diff
@@ -0,0 +1,124 @@
...
+        mlp_only: ["*.mlp.*"]
+        attention_only: ["*.self_attn.*"]
+      exclude_patterns:
+        all_but_layer_0: ["model.layers.0.*"]
+        all_but_layer_1: ["model.layers.1.*"]
+        all_but_layer_2: ["model.layers.2.*"]
+        all_but_layer_3: ["model.layers.3.*"]
+  - classname: "GeometricSimilarityComparison"
+    extra_init_kwargs:
+      reference_run_path: "wandb:goodfire/spd/runs/2js1ccon"
```
danbraunai, line 87: immediate thoughts: it strikes me as off having a reference run path in the config for a new run. The standard use case in my head after reading the PR description is to having a post-hoc script which does the comparisons when the runs have finished (or on intermediate checkpoints). The issue there is waiting for runs to finish (or wasting space with saving intermediate checkpoints).

Will think more when reviewing and afterwards. Interested in thoughts.

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
 
+class GeometricSimilarityComparison(StreamingEval):
+    SLOW = True
+
+    def __init__(self, model: ComponentModel, config: Config, **kwargs: Any):
+        self.model = model
+        self.config = config
+        self.reference_run_path = kwargs.get("reference_run_path")
+        if self.reference_run_path is None:
+            raise ValueError("reference_run_path is required for GeometricSimilarityComparison")
```
danbraunai, line 767: Perhaps you want the below?
```suggestion
    def __init__(self, model: ComponentModel, config: Config, reference_run_path: str, **kwargs: Any):
        self.model = model
        self.config = config
        self.reference_run_path = reference_run_path
```

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+    def compute(self) -> Mapping[str, float]:
+        """Compute the geometric similarity metrics"""
+
+        activation_densities = {
+            module_name: self.component_activation_counts[module_name] / self.n_tokens
+            for module_name in self.model.components
+        }
+
+        if self._computed_this_eval:
+            return {}
```
danbraunai, line 907: Would this ever be true? We only run `compute()` at the end of an eval. Other evals don't have this pattern  (maybe they should have something like this, but I don't like only having it in this eval).

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+
+        try:
+            similarities = self._compute_subcomponent_geometric_similarities(activation_densities)
+            self._computed_this_eval = True
+            return similarities
+        except Exception as e:
```
danbraunai, line 913: I'd only have this try/except if it's possible for it to fail late in a run. If it will fail at step 0, I'd prefer to just have it raise an error immediately and let the user update it.

Relatedly, if keeping this, I think you should catch the specific exceptions you expect to see.

spd/registry.py
danbraunai, line 1: Related to the config comment, if we remove the configs we'd remove these from the registry.

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+        self.n_tokens = 0
+        self.component_activation_counts: dict[str, Float[Tensor, " C"]] = {
+            module_name: torch.zeros(model.C, device=self.device)
+            for module_name in model.components
+        }
+
+    def _load_reference_model(self) -> ComponentModel:
+        """Load the reference model from wandb or local path"""
+        if self.reference_model is None:
+            from spd.models.component_model import ComponentModel
```
danbraunai, line 781: This is already imported at the top of the file. I'm surprised the linter doesn't complain here.

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+
+            assert self.reference_run_path is not None, (
+                "reference_run_path should not be None at this point"
+            )
```
danbraunai, line 785: I'd remove the comment here, doesn't add anything and takes up 2 lines

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+        self.reference_model: ComponentModel | None = None
+        self._computed_this_eval = False
+        self.device = next(iter(model.parameters())).device
+        self.n_tokens = 0
+        self.component_activation_counts: dict[str, Float[Tensor, " C"]] = {
+            module_name: torch.zeros(model.C, device=self.device)
+            for module_name in model.components
+        }
+
+    def _load_reference_model(self) -> ComponentModel:
```
danbraunai, line 778: Another reason that I don't like doing this comparison during a run in the same process is because loading a reference model could be memory intensive and might make memory management during training more spiky (specifically, it may increase the amount of total gpu memory required for the run just to compute this eval).

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+
+            assert self.reference_run_path is not None, (
+                "reference_run_path should not be None at this point"
+            )
+            self.reference_model = ComponentModel.from_pretrained(self.reference_run_path)
+
+            if torch.cuda.is_available():
+                self.reference_model.to("cuda")
+                self.reference_model.eval()
+                self.reference_model.requires_grad_(False)
```
danbraunai, line 791: 1. I'd instead add `self.device = next(iter(model.parameters())).device` to the class init and then do `model.to(device)`. (see other evals)
2. You only put it in eval mode and set requires_grad_(False) when cuda is available, which is odd.

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+class GeometricSimilarityComparison(StreamingEval):
+    SLOW = True
+
+    def __init__(self, model: ComponentModel, config: Config, **kwargs: Any):
+        self.model = model
+        self.config = config
+        self.reference_run_path = kwargs.get("reference_run_path")
+        if self.reference_run_path is None:
+            raise ValueError("reference_run_path is required for GeometricSimilarityComparison")
+        self.kwargs = kwargs
```
danbraunai, line 768: I don't like this. I'd specifically save the things you need. E.g. density threshold

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+
+            # Extract U and V matrices
+            C = current_components.C
+            current_U = current_components.U  # Shape: [C, d_out]
+            current_V = current_components.V  # Shape: [d_in, C]
+            ref_U = reference_components.U
+            ref_V = reference_components.V
+
+            # Throw away components that are not active enough in the current model
+            density_threshold = self.kwargs.get("density_threshold", 0.0)
```
danbraunai, line 826: See my comment above about kwargs, I'd specifically save the density_threshold rather than save "kwargs". I'm also unsure about having this default value of 0. Do we need a default? Can't we force the user to define a density threshold? Defaults are scary, especially when hidden away so deep in the stack trace.

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+            C_alive = sum(activation_densities[layer_name] > density_threshold)
+            if C_alive == 0:
+                logger.warning(
+                    f"\n WARNING:No components are active enough in {layer_name} for density threshold {density_threshold}. Geometric similarity comparison failed to run. \n"
```
danbraunai, line 830: Split over multiple lines

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+                )
+                continue
+            current_V = current_V[:, activation_densities[layer_name] > density_threshold]
+            current_U = current_U[activation_densities[layer_name] > density_threshold]
```
danbraunai, line 834: I'd use a different name to current_V and current_U since this is a different thing to the original current_V and current_U

spd/eval.py
```diff
@@ -755,6 +756,165 @@ def compute(self) -> Mapping[str, float]:
...
+
+            # Find max abs cosine similarity for each current component
+            max_similarities = cosine_sim_matrix.max(dim=1).values
+            similarities[f"mean_max_abs_cosine_sim/{layer_name}"] = max_similarities.mean().item()
+            similarities[f"max_abs_cosine_sim_std/{layer_name}"] = max_similarities.std().item()
+            similarities[f"max_abs_cosine_sim_min/{layer_name}"] = max_similarities.min().item()
+            similarities[f"max_abs_cosine_sim_max/{layer_name}"] = max_similarities.max().item()
+
+        # Compute a metrics across all model components for each type of metric
+        # First get the metric names by stripping away the layer name
```
danbraunai, line 865: I'd remove these comments

danbraunai (changes requested):
Thanks! Main comment is that I don't think it makes sense to have this setup where you put in a reference run in a run config and then it produces the plots in wandb. Instead, I think it makes more sense to have a separate script which takes in two runs and computes this metrics.

This seems like the type of analysis that we'll just want to run sporadically by comparing two runs with similar seeds/whatever else. It doesn't seem like the type of analysis that we want on every run. It also brings in a potentially heavy memory footprint.

spd/eval.py
danbraunai: (guess this is no longer relevant if making this a script rather than an eval. But what I mean was to catch e.g. ValueError instead of any Exception.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+        if task_name == "tms":
+            from spd.experiments.tms.models import TMSTargetRunInfo
+            from spd.utils.data_utils import DatasetGeneratedDataLoader, SparseFeatureDataset
+
+            if "pretrained_model_path" not in config or not config["pretrained_model_path"]:
+                raise ValueError("pretrained_model_path must be set for TMS models")
+
+            target_run_info = TMSTargetRunInfo.from_path(config["pretrained_model_path"])
+            n_features = target_run_info.config.tms_model_config.n_features
+            synced_inputs = target_run_info.config.synced_inputs
```
danbraunai, line 95: I wouldn't bother creating these new variables if they're just used once and you can call them directly without adding new lines.

danbraunai: I noticed these elsewhere in this PR, worth doing a read over.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+            from spd.experiments.ih.model import InductionModelTargetRunInfo
+            from spd.utils.data_utils import DatasetGeneratedDataLoader, InductionDataset
+
+            if "pretrained_model_path" not in config or not config["pretrained_model_path"]:
+                raise ValueError("pretrained_model_path must be set for Induction Heads models")
+
+            target_run_info = InductionModelTargetRunInfo.from_path(config["pretrained_model_path"])
+            vocab_size = target_run_info.config.ih_model_config.vocab_size
+            seq_len = target_run_info.config.ih_model_config.seq_len
+            prefix_window = task_config.get("prefix_window") or seq_len - 3
```
danbraunai, line 182: same here RE removing new variables that are only used once.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+
+        raise ValueError(
+            f"Unsupported task type: {task_name}. Supported types: tms, lm, resid_mlp, ih"
+        )
+
+    def compute_activation_densities(
+        self, model: ComponentModel, eval_iterator: Iterator[Any], n_steps: int = 5
+    ) -> dict[str, Float[Tensor, " C"]]:
+        """Compute activation densities using same logic as ComponentActivationDensity."""
+        # Get config for this model
```
danbraunai, line 206: remove comment

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+                    shuffle=self.comparison_config.get("shuffle_data", False),
+                )
+            )
+
+        raise ValueError(
+            f"Unsupported task type: {task_name}. Supported types: tms, lm, resid_mlp, ih"
+        )
+
+    def compute_activation_densities(
+        self, model: ComponentModel, eval_iterator: Iterator[Any], n_steps: int = 5
```
danbraunai, line 203: I'd remove the default n_steps: int = 5. I don't see the benefit when this isn't a public function that people will use from a lot of places.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+                similarities[f"{metric_name}/all_layers"] = sum(values) / len(values)
+
+        return similarities
+
+    def run_comparison(
+        self, eval_iterator: Iterator[Any], n_steps: int | None = None
+    ) -> dict[str, float]:
+        """Run the full comparison pipeline."""
+        if n_steps is None:
+            n_steps = self.comparison_config.get("n_eval_steps", 5)
```
danbraunai, line 336: this default pattern is a bit scary, but maybe it's OK if you think it is useful for this specific script.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+        assert isinstance(n_steps, int)  # Ensure n_steps is an int for type checking
```
danbraunai, line 337: Can remove comment

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+        help="Path to configuration file",
+    )
+    args = parser.parse_args()
+
+    config = load_config(args.config, dict)
+    current_model_path = config["current_model_path"]
+    reference_model_path = config["reference_model_path"]
+    density_threshold = config.get("density_threshold", 0.0)
+    device = config.get("device", "auto")
+    output_dir = Path(config.get("output_dir", "./comparison_results"))
```
danbraunai, line 366: I think we should write results in an "out" directory that is relative to this specific file (`Path(__file__) / "out"`). As per my earlier comment which says that this compare models should be in its own directory within scripts, the outputs would then go in an out directory that's specific to the compare_models experiments. As it stands, this will write results in directories relative to wherever the user invokes this script. This is going to put them in their git working directory if they run it from the repo somewhere, and might put them in different paths if they invoke the script from different places.

danbraunai (changes requested):
Looks good. Mostly minor comments.

Some comments I made at specific locations but refer to code in several places. E.g. that I think `dict.get` should probably be avoided everywhere in this script, and relatedly, that defaults should be avoided.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+                data_generation_type=task_config["data_generation_type"],
+                value_range=(0.0, 1.0),
+                synced_inputs=synced_inputs,
+            )
+            return iter(
+                DatasetGeneratedDataLoader(
+                    dataset,
+                    batch_size=self.comparison_config.get(
+                        "eval_batch_size", 1
+                    ),  # TODO get rid of 'get' pattern
```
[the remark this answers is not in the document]
leesharkey: Yeah agree on both counts. My bad for missing my own TODO... Thanks for flagging.

spd/scripts/compare_models.py
```diff
@@ -0,0 +1,393 @@
...
+        """Create evaluation data loader using exact same patterns as decomposition scripts."""
+        task_config = config.get("task_config", {})
+        task_name = task_config.get("task_name")
+
+        if not task_name:
+            raise ValueError("task_config.task_name must be set")
+
+        if task_name == "tms":
+            from spd.experiments.tms.models import TMSTargetRunInfo
+            from spd.utils.data_utils import DatasetGeneratedDataLoader, SparseFeatureDataset
```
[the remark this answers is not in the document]
leesharkey: uh, yeah 😅 Okay I'll not take action on this for now. If these function end up making their way into core files then I'll change.

PR #158, opened 2025-09-18
Hidden activation reconstruction loss

leesharkey:
## Description
This PR implements hidden activation reconstruction, which allows SPD runs either to monitor or potentially optimize for hidden activation reconstruction losses.

The main thing it adds is a "pre_forward_cache_components" mode component_model.py. This caches the inputs to each SPD module. It does for "component" mode what "pre_forward_cache" does for "target mode. 

It then also modifies the stochastic recon forward pass also to return hidden activations if necessary. Most of the other changes are changes to config files and tests. 

## Related Issue
None

## Motivation and Context
Main motivation: We might be able to identify pathological runs earlier if we are tracking hidden activation reconstruction. We might have more quickly noticed SPD using the last layer to cheat, for instance. 
Secondary motivation: We might get some benefits for weakly optimizing for hidden layer activation reconstruction loss as an auxiliary loss. It might make it harder for networks to cheat in general, for instance. But it may also reconstruct unwanted noise, so we wouldn't want to optimize for it too strongly.

## How Has This Been Tested?
The losses run and look sensible on a representative sample of models:
- TMS 5-2
- Resid-mlp 2 layer
- ss_llama

No.

### DIFF
--- a/spd/losses.py
+++ b/spd/losses.py
@@ -155,7 +155,6 @@ def calc_masked_recon_loss(
             stochastic sources).
         target_out: Target model output
         output_recon_loss_type: Type of loss to calculate for output reconstruction
-        device: Device to run computations on
 
     Returns:
         The recon loss


spd/models/component_model.py
```diff
@@ -510,6 +522,50 @@ def cache_hook(_: nn.Module, input: tuple[Tensor, "..."], param_name: str) -> No
...
+        # Set up component replacements
+        with self._replaced_modules(mask_infos=mask_infos):
+            try:
+                raw_out = self.patched_model(*args, **kwargs)
+                out = self._extract_output(raw_out)
+                return out, cache
+            finally:
+                for handle in handles:
+                    handle.remove()
+
```
danbraunai, line 568: I think you can just directly use _forward_with_pre_forward_cache_hooks with minor modifications instead of creating this new method? Modifications:
- You can add a "mask_infos" argument to it, which can be None.
- Add the `with self._replaced_modules` context manager to the method. Pass in an empty dictionary when mask_infos is None. It should then be a no-op method (apart from calling make_pristine(), which is fine).

Everything else should be the same unless I'm missing something.

spd/losses.py
```diff
@@ -184,25 +186,63 @@ def calc_masked_recon_loss(
...
+
+    output_loss = total_out_recon_loss / len(mask_infos_list)
+    if return_hidden_act_recon_losses and target_hidden is not None:
+        for layer_name in hidden_losses:
+            hidden_losses[layer_name] = hidden_losses[layer_name] / len(mask_infos_list)
+
+    if return_hidden_act_recon_losses:
+        return output_loss, hidden_losses
+    else:
+        return output_loss
```
danbraunai, line 245: I have to run now and can't look at the best alternatives, but we should always avoid a function having two different return types. Makes things very hard to handle (especially typing). At a minimum you can do an `@overload` thing where you write out the two different function signatures that you expect (like [here](https://www.codementor.io/@arpitbhayani/overload-functions-in-python-13e32ahzqt#function-overloading-in-action)). But I'd much prefer to just refactor things to avoid this. E.g. making a new loss function entirely. If they share a lot of code then trying to make that code a separate function which both of them call.

spd/losses.py
```diff
@@ -318,15 +361,36 @@ def calculate_losses(
             )
             stoch_mask_infos_list.append(stoch_mask_infos)
 
-        stochastic_recon_loss = calc_masked_recon_loss(
+        # Check if we also need to compute hidden activation losses
```
danbraunai, line 364: Remove comment

spd/losses.py
```diff
@@ -318,15 +361,36 @@ def calculate_losses(
...
+        compute_hidden_losses = (
+            config.hidden_act_recon_coeff is not None and target_hidden is not None
+        )
+
+        # Call calc_masked_recon_loss with appropriate parameters
```
danbraunai, line 369: Remove comment

spd/experiments/ih/ih_config.yaml
```diff
@@ -29,7 +29,8 @@ recon_layerwise_coeff: null
 stochastic_recon_layerwise_coeff: 1
 importance_minimality_coeff: 1e-2
 pnorm: 0.1
-output_loss_type: kl
+output_recon_loss_type: kl
+hidden_act_recon_coeff: 0.0
```
danbraunai, line 33: I want to rewrite our losses and evals so that we don't have to do this. If we want to track something but not compute it in our loss, we'd just run it in our evals and not on every training step. I'll be looking to make a PR to address this in the coming days. For now I guess it's OK to do this provided adding the hidden reconstruction calculation every step doesn't meaningfully increase runtime or memory consumption (e.g. by more than 5%).

spd/losses.py
danbraunai: Just had a look at the logic. I think the hidden loss can just be in its own loss function. Unless I'm missing something, I don't see any real computational savings from having these two in the same function.

danbraunai: @leesharkey I chatted to AI about this. Here's a solution that is probably fine for now. You call the run_masked_forward if either of the loss coefficients are not None. Note that you'll want to update the names of many of these variables.

```
from dataclasses import dataclass
from typing import Iterable
from jaxtyping import Float
import torch
from torch import Tensor

@dataclass(frozen=True)
class MaskedForwardCache:
    outputs: list[Float[Tensor, "... d_model_out"]]
    hidden_by_layer: dict[str, list[Float[Tensor, "..."]]]  # only present for requested layers

def run_masked_forward(
    model: ComponentModel,
    batch: Float[Tensor, "... d_in"],
    mask_infos_list: list[dict[str, ComponentsMaskInfo]],
    device: str,
    module_names: Iterable[str] | None = None,
) -> MaskedForwardCache:
    outputs: list[Tensor] = []
    hidden: dict[str, list[Tensor]] = {m: [] for m in (module_names or [])}
    for mask_infos in mask_infos_list:
        if module_names:
            out, acts = model(batch, mode="pre_forward_cache_components",
                              mask_infos=mask_infos, module_names=list(module_names))
            for name in hidden:
                hidden[name].append(acts[name])
        else:
            out = model(batch, mode="components", mask_infos=mask_infos)
        outputs.append(out)
    return MaskedForwardCache(outputs=outputs, hidden_by_layer=hidden)

def output_recon_loss_from_cache(
    cache: MaskedForwardCache,
    target_out: Float[Tensor, "... d_model_out"],
    kind: Literal["mse", "kl"],
) -> Float[Tensor, ""]:
    if kind == "mse":
        return torch.stack([((o - target_out) ** 2).mean() for o in cache.outputs]).mean()
    return torch.stack([calc_kl_divergence_lm(pred=o, target=target_out) for o in cache.outputs]).mean()

def hidden_recon_losses_from_cache(
    cache: MaskedForwardCache,
    target_hidden: dict[str, Tensor],
) -> dict[str, Float[Tensor, ""]]:
    return {
        layer: torch.stack([((a - target_hidden[layer]) ** 2).mean() for a in acts]).mean()
        for layer, acts in cache.hidden_by_layer.items()
    }
```

I am likely to restructure things when I work out how to manage the training losses and eval metrics. But this should be good for now.

spd/losses.py
```diff
@@ -318,15 +396,37 @@ def calculate_losses(
...
+                total_loss += config.hidden_act_recon_coeff * layer_loss
+                loss_terms[f"hidden_act_recon/{layer_name}"] = layer_loss.item()
+        else:
+            stochastic_recon_loss = calc_masked_recon_loss(
+                model=model,
+                batch=batch,
+                mask_infos_list=stoch_mask_infos_list,
+                target_out=target_out,
+                output_recon_loss_type=config.output_recon_loss_type,
+            )
```
danbraunai, line 428: I think you can simplify this to:
```
        cache = run_masked_forward(
            model=model,
            batch=batch,
            mask_infos_list=stoch_mask_infos_list,
            hidden_module_names=target_hidden.keys() if config.hidden_act_recon_coeff else None,
        )
        stochastic_recon_loss = output_recon_loss_from_cache(cache, target_out, config.output_recon_loss_type)
        if config.hidden_act_recon_coeff:
            assert config.hidden_act_recon_coeff is not None
            hidden_losses = hidden_recon_losses_from_cache(cache, target_hidden)
            for layer_name, layer_loss in hidden_losses.items():
                total_loss += config.hidden_act_recon_coeff * layer_loss
                loss_terms[f"hidden_act_recon/{layer_name}"] = layer_loss.item()
```
(this code assumes that target_hidden is always not None, which I think it should be?

I'd ideally further simplify the loss calculations to not have two different functions which both do the mse calculation of `torch.stack([((o - target_out) ** 2).mean() for o in ...]).mean()`. But I might manage this in a future PR/refactor if/when I get around to it.

danbraunai (changes requested):
nice, liking this. Few more suggestion improvements then I think we'll be good to go. Good to hear that the timecost is trivial.

spd/models/component_model.py
```diff
@@ -377,7 +377,8 @@ def _make_gates(
     def forward(
         self,
         *args: Any,
-        mode: Literal["target", "components", "pre_forward_cache"] | None = "target",
+        mode: Literal["target", "components", "pre_forward_cache", "pre_forward_cache_components"]
```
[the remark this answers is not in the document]
leesharkey: As per Dan and my IRL discussion, I'll ignore this, since Dan was confused about this and thought pre_forward_cache_components was now redundant when in fact it is not.

spd/losses.py
leesharkey: RE "this code assumes that target_hidden is always not None, which I think it should be?", see discussion above (https://github.com/goodfire-ai/spd/pull/158#discussion_r2369079732)

PR #162, opened 2025-09-20
Consolidate losses and evals

danbraunai:

### DIFF
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -37,7 +37,7 @@ class EvalMetricConfig(BaseModel):
     )
 
     def _get_metric_class(self) -> type | None:
-        available_classes = importlib.import_module("spd.eval").EVAL_CLASSES
+        available_classes = importlib.import_module("spd.metrics").EVAL_CLASSES
         cls = available_classes.get(self.classname)
         if cls is None:
             logger.warning(
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -25,7 +25,7 @@ from spd.models.components import GateType
 from spd.spd_types import ModelPath, Probability
 
 
-class EvalMetricConfig(BaseModel):
+class MetricConfig(BaseModel):
     model_config: ClassVar[ConfigDict] = ConfigDict(extra="forbid", frozen=True)
     classname: str = Field(
         ...,
@@ -34,6 +34,17 @@ class EvalMetricConfig(BaseModel):
     extra_init_kwargs: dict[str, Any] = Field(
         default={},
         description="Extra keyword arguments to pass to the class constructor besides `model: ComponentModel` and `config: Config`",
+    )
+    coeff: float | None = Field(
+        default=None,
+        description="Optional coefficient used for weighting into loss/total; if None, uses the corresponding *_coeff field on Config.",
+    )
+    slow: bool | None = Field(
+        default=None,
+        description=(
+            "Optional override of the metric's SLOW flag for gating in evaluation. "
+            "If None, uses the class-level SLOW on the metric."
+        ),
     )
 
     def _get_metric_class(self) -> type | None:
@@ -261,9 +272,16 @@ class Config(BaseModel):
         description="Interval (in steps) at which to save model checkpoints (None disables saving "
         "until the end of training).",
     )
-    eval_metrics: list[EvalMetricConfig] = Field(
+    eval_metrics: list[MetricConfig] = Field(
         default=[],
         description="List of metrics to use for evaluation",
+    )
+    loss_metrics: list[MetricConfig] = Field(
+        default=[],
+        description=(
+            "List of loss metrics to compute (used for both training logs and eval); "
+            "coefficients provided here are also used for weighting the training loss and eval loss/total."
+        ),
     )
     include_loss_metrics_in_eval: bool = Field(
         default=True,
@@ -325,6 +343,7 @@ class Config(BaseModel):
         "image_freq",
         "metrics_fns",
         "figures_fns",
+        "include_loss_metrics_in_eval",
     ]
     RENAMED_CONFIG_KEYS: ClassVar[dict[str, str]] = {
         "print_freq": "eval_freq",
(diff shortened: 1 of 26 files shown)


spd/registry.py
danbraunai, line 1: Unrelated, just cleaning up

spd/run_spd.py
```diff
@@ -37,17 +35,14 @@
...
     get_lr_with_warmup,
 )
 from spd.utils.module_utils import replace_std_values_in_layernorm
 from spd.utils.run_utils import save_file
 
 
-def local_log(
-    data: Mapping[str, float | Image.Image | wandb.plot.CustomChart], step: int, out_dir: Path
-) -> None:
+def local_log(data: dict[str, Any], step: int, out_dir: Path) -> None:
```
danbraunai, line 45: The data values will raise errors if they're not json serializable or not handled specifically in the function.

oli-clive-griffin:
Quickly:
> Added https://github.com/goodfire-ai/spd/issues/176 to convert AliveComponentsTracker to a Metric

This is awesome.

spd/configs.py
```diff
@@ -22,55 +19,182 @@
 from spd.experiments.tms.configs import TMSTaskConfig
 from spd.log import logger
 from spd.models.components import GateType
+from spd.models.sigmoids import SigmoidTypes
 from spd.spd_types import ModelPath, Probability
 
+#### Train Metric Configs ####
+TrainMetricClassname = Literal[
```
oli-clive-griffin, line 26: this is unused AFAICT

spd/configs.py
```diff
@@ -22,55 +19,182 @@
...
+    "StochasticReconSubsetCEAndKL",
+]
+
+
+class EvalMetricConfig(BaseModel):
+    model_config: ClassVar[ConfigDict] = ConfigDict(extra="forbid", frozen=True)
+
+    slow: bool = Field(
+        default=False,
+        description=("If True, only run this metric once every slow_eval_freq steps."),
```
oli-clive-griffin, line 102: Do you want this to be configurable? My assumption is no, in which case could we just make it a static `ClassVar`?

oli-clive-griffin: (goes for all other slow metrics)

oli-clive-griffin: actually - Why not just put it on the metric computation class, then skip by `metric.slow` (as a static classvar again)

spd/configs.py
```diff
@@ -326,22 +410,63 @@ def microbatch_size(self) -> PositiveInt:
...
+        loss_metric_configs. This is done by DEPRECATED_CONFIG_KEYS. This means that old runs which
+        have these keys will not be able to be used for e.g. finetuning, but the raw model will be
+        the same.
+        """
+
+        # We don't bother doing the complex mapping from an old structure with the ``eval_metrics``
+        # key to the new structure with the ``eval_metric_configs`` key.
+        config_dict.pop("eval_metrics", None)
+
+        # Remove SubsetReconstructionLoss if it appears
```
oli-clive-griffin, line 446: Why do we specifically need to support this one? seems like surely there shouldn't already be deprecated `Metric`s before this PR is even merged. Am I missing something?

spd/configs.py
```diff
@@ -357,15 +482,6 @@ def handle_deprecated_config_keys(cls, config_dict: dict[str, Any]) -> dict[str,
...
     def validate_model(self) -> Self:
-        # If any of the coeffs are 0, raise a warning
-        msg = "is 0, you may wish to instead set it to null to avoid calculating the loss"
-        if self.ci_recon_coeff == 0:
-            logger.warning(f"recon_coeff {msg}")
-        if self.importance_minimality_coeff == 0:
-            logger.warning(f"importance_minimality_coeff {msg}")
-        if self.faithfulness_coeff == 0:
-            logger.warning(f"faithfulness_coeff {msg}")
-
```
oli-clive-griffin, line 368: 😍

spd/losses.py
```diff
@@ -4,295 +4,128 @@
...
-    total_loss += config.importance_minimality_coeff * importance_minimality_loss
-    loss_terms["importance_minimality"] = importance_minimality_loss.item()
-
-    loss_terms["total"] = total_loss.item()
-
-    return total_loss, loss_terms
+    total = torch.tensor(0.0, device=batch.device)
+    terms: dict[str, float] = {}
+
+    for cfg in loss_metric_configs:
```
oli-clive-griffin, line 52: I have an idea here, noting as reminder for myself

spd/models/component_model.py
```diff
@@ -107,12 +107,11 @@ def __init__(
         self.target_model = target_model
         self.C = C
         self.pretrained_model_output_attr = pretrained_model_output_attr
-
-        module_paths = get_target_module_paths(target_model, target_module_patterns)
+        self.module_paths = get_target_module_paths(target_model, target_module_patterns)
```
oli-clive-griffin, line 110: nice, does this replace `list(component_model.components.keys())` throughout the codebase?

spd/configs.py
danbraunai: I did have in mind for it to be configurable. But now that you say it, I suppose people can just hardcode class variable if they want to change it for an eval they're particularly interested in sometimes, that's not a big deal.

spd/models/component_model.py
danbraunai: Oh yeah that's nice. Just pushed a change with this.

spd/configs.py
danbraunai: Removed.

spd/configs.py
danbraunai: Now it's only in classvars.

spd/metrics/base.py
```diff
@@ -1,194 +1,36 @@
...
-    3. Implement `compute()` to calculate final metric values
-
-    For distributed evaluation:
-    - Call `sync_dist()` before `compute()` to synchronize states across ranks
-    - This gives you explicit control over when synchronization happens
-    """
+    """Interface for metrics that can be used in training and evaluation."""
 
-    is_differentiable: bool | None = None
+    slow: bool = False
```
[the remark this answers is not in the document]
oli-clive-griffin: (I think that's how you do it)

spd/metrics/README.md
oli-clive-griffin, line 1: just noting that the modifications my branch had to this file were completely vibe-coded and not checked

spd/scripts/compare_models/compare_models.py
```diff
@@ -253,7 +253,7 @@ def compute_activation_densities(
                 batch = extract_batch_data(next(eval_iterator))
                 batch = batch.to(self.device)
                 _, pre_weight_acts = model(
-                    batch, mode="pre_forward_cache", module_names=list(model.components.keys())
+                    batch, mode="pre_forward_cache", module_names=model.module_paths
```
oli-clive-griffin, line 256: thoughts on making module_names=model.module_paths the default behaviour?

```
def forward(
   ...
   module_names: list[str] | None
   ...
)
   module_names = module_names or self.module_paths
```

oli-clive-griffin (approved):
Nice. love how this has turned out, feel free to merge after addressing

oli-clive-griffin:
> I don't love the current structure of our metric configs:
> 1. Most of them just have a classname and nothing else
> 2. The classname itself corresponds to the name of the actual Metric class, but it could be anything since we just use it as a discriminated union.
> But I can't immediately think of a cleaner option. Note that even though they don't have extra params in the config, some are initialized with different arguments.

I honestly think this is completely fine.

spd/metrics/README.md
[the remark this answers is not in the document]
danbraunai: Agreed, I'm just deleting this file. There's a lot of examples for people to read, and everything is explicit enough now.

spd/utils/distributed_utils.py
```diff
@@ -98,8 +98,8 @@ def init_distributed(backend: Literal["nccl", "gloo"] | None = None) -> Distribu
             device_id=local_device,
         )
 
-    # Set the default cuda device for this process
-    if torch.cuda.is_available():
+    # Set the default cuda device for this process (only for NCCL backend)
+    if backend == "nccl" and torch.cuda.is_available():
```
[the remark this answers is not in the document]
danbraunai: I've removed this. But, the reason it was there:

For some distributed tests we use gloo and two processes (mpirun -np 2 ...). If we run these tests on a machine that has a gpu available, then, unless we also guard this by `backend == "nccl"`, it will set the cuda device with some non-zero rank. This was an issue because, when running tests on a machine with one gpu but with multiple processes, it would try to use different gpu device ids.

I've fixed this by just forcing the device to be "cpu" in distributed_utils.get_device() when backend == gloo so it never tries to use a gpu.

oli-clive-griffin (approved):
Beautiful 👨‍🍳🤌

PR #165, opened 2025-09-22
Refactor to use hooks instead of `ComponentsOrModule`

oclivegriffin:
## Description
A couple of small changes first:
- `GateMLPs`, `VectorGateMLPs`, and `LayerwiseGlobalMLP` renamed to `MLPGates`, `VectorMLPGates`, and `SharedVectorMLPGates`
- all usages of "original" renamed to "target"
- "pre_forward_cache" term now called "input_cache"

The main change is that instead of mutating the module structure to wrap target modules in `ComponentsOrModule`, we now simply keep a ModuleDict of components on `ComponentModel`, and use hooks to intervene on the forward pass.

Here's an attempt to draw what happens if you add a pre identity to a linear module and then decompose both the identity and the module, on a given forward pass

image

## Motivation and Context
Sets the stage for subset reconstruction loss, vastly cleans up the component handling and identity insertion code.

## How Has This Been Tested?
Have added I think a reasonable amount of testing. Also doing some runs [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250922_165457--VmlldzoxNDQ4ODEyNg==)

Don't believe so. all canonical runs load

### DIFF
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -126,6 +126,7 @@ class Config(BaseModel):
         "matrix should be inserted and decomposed beforehand",
     )
 
+    @property
     def all_module_patterns(self):
         if self.identity_module_patterns is None:
             return self.target_module_patterns
--- a/spd/models/component_model.py
+++ b/spd/models/component_model.py
@@ -446,7 +446,7 @@ class ComponentModel(LoadableModule):
 
         comp_model = ComponentModel(
             target_model=target_model,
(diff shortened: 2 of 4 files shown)


oli-clive-griffin:
test runs happening [here](https://wandb.ai/goodfire/spd/reports/SPD-Run-Report---run_20250922_165457--VmlldzoxNDQ4ODEyNg==)

PR #168, opened 2025-09-23
Routing / Subset recon loss

oclivegriffin:
## Description
Adds "subset reconstruction loss". In order to do so, introduces the concept of "routing": For a module where we're using components for the forward pass, the routing defines, for each position (either `(batch,)` or `(batch, seq)`), the set of modules in which the components are used in place of the target weights.

This allows us to enable and disable different **subsets** of modules for different positions. Like this:

image

We use this to implement subset recon loss by passing in a routing mask. specifically, we use what I've called "uniform-k stochastic" sampling:
- for each position, draw an integer `k` uniformly from (1, n_modules)
- given that k value, randomly activate k random modules for that position

I've included 2 variants, `ci_recon_subset_coeff` (using CI as masks) and `stochastic_recon_subset_coeff` (using stochastic masks). TODO: change name to `ci_masked_recon_subset_coeff`.

## Motivation and Context
to enable subset recon (@danbraunai idk if I'm missing something but I feel like we could remove this section of the PR template. I always end up explaining what would go in here in the description)

## How Has This Been Tested?
Added a test, but I think the implementation is pretty straight forward tbh. I'll kick off some runs that use it:

https://wandb.ai/goodfire/spd/reports/Routing-Smoke-Test--VmlldzoxNDUwMzA0Mg
Nope

### DIFF
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -161,7 +161,7 @@ class Config(BaseModel):
         description="Coefficient for recon loss with stochastically sampled masks on one layer at "
         "a time",
     )
-    ci_recon_subset_coeff: NonNegativeFloat | None = Field(
+    ci_masked_recon_subset_coeff: NonNegativeFloat | None = Field(
         default=None,
         description="Coefficient for recon loss with causal importance mask and routed components",
     )
--- a/spd/losses.py
+++ b/spd/losses.py
@@ -241,7 +241,7 @@ def calculate_losses(
         loss_terms["stochastic_recon_layerwise"] = stochastic_recon_layerwise_loss.item()
 
     # CI subset reconstruction loss
-    if config.ci_recon_subset_coeff is not None:
+    if config.ci_masked_recon_subset_coeff is not None:
         subset_routing_masks = sample_uniform_k_subset_routing_masks(
             mask_shape=next(iter(causal_importances.values())).shape[:-1],
             modules=list(causal_importances.keys()),
@@ -260,7 +260,7 @@ def calculate_losses(
             loss_type=config.output_loss_type,
             device=device,
         )
-        total_loss += config.ci_recon_subset_coeff * ci_recon_subset_loss
+        total_loss += config.ci_masked_recon_subset_coeff * ci_recon_subset_loss
         loss_terms["ci_recon_subset"] = ci_recon_subset_loss.item()
 
     # Stochastic reconstruction subset loss


spd/models/component_model.py
```diff
@@ -335,20 +335,23 @@ def fwd_hook(
             _module: nn.Module,
             args: list[Any],
             kwargs: dict[Any, Any],
-            _output: Any,
+            output: Any,
```
danbraunai, line 338:
```suggestion
            output: Float[Tensor, "... d_out"],
```

spd/models/component_model.py
```diff
@@ -335,20 +335,23 @@ def fwd_hook(
...
             components: Components,
             mask_info: ComponentsMaskInfo,
         ) -> None | Any:
             assert len(args) == 1, "Expected 1 argument"
             assert len(kwargs) == 0, "Expected no keyword arguments"
             x = args[0]
             assert isinstance(x, Tensor), "Expected input tensor"
+            assert isinstance(output, Tensor), "Expected output tensor"
```
danbraunai, line 346:
```suggestion
            assert isinstance(output, Tensor), "Only supports single-tensor outputs, got type(output)"
```

spd/models/components.py
```diff
@@ -300,13 +299,17 @@ class ComponentsMaskInfo:
...
+    weight_delta_and_mask: WeightDeltaAndMask | None = None
 
 
 def make_mask_infos(
-    component_masks: Mapping[str, Float[Tensor, "... C"]],
+    component_masks: dict[str, Float[Tensor, "... C"]],
+    routing_masks: dict[str, Bool[Tensor, "..."]] | None = None,
     weight_deltas_and_masks: dict[str, WeightDeltaAndMask] | None = None,
 ) -> dict[str, ComponentsMaskInfo]:
     """Create ComponentsMaskInfo dict from dicts of component masks, and weight deltas and weight delta masks.
```
danbraunai, line 315:
```suggestion
    """Create ComponentsMaskInfo dict from dicts of component masks, weight deltas and weight delta masks.
```

danbraunai: This docstring is also missing an argument doc for routing_mask

spd/utils/component_utils.py
```diff
@@ -22,24 +22,87 @@ def _sample_stochastic_mask(
...
+    # turn values into ranks via double argsort trick. (for example: [0.8, 0.2, 0.3] -> [2, 0, 1])
+    return noise.argsort(dim=dim).argsort(dim=dim)
+
+
+def sample_uniform_k_subset_routing_masks(
+    mask_shape: tuple[int, ...],
+    modules: list[str],
+    device: torch.device | str,
+) -> dict[str, Bool[Tensor, "..."]]:
+    """Creates routing masks for each module such that the number of modules routed to for each position is independent and uniformly sampled from [1, len(modules)]
```
danbraunai, line 46: Wrap to multiple lines

spd/utils/component_utils.py
```diff
@@ -22,24 +22,87 @@ def _sample_stochastic_mask(
...
+
+    Achieves this by:
+    - for each position, k is independent and uniformly sampled from [1, len(modules)]
+    - for each position, a k-sized random subset of modules are routed to
+
+    Args:
+        mask_shape: Shape of the routing masks, likely (batch,) or (batch, seq_len)
+        modules: List of modules to route to
+    """
```
danbraunai, line 55: Add a returns section in the docstring.

spd/configs.py
```diff
@@ -161,6 +161,14 @@ def all_module_patterns(self):
         description="Coefficient for recon loss with stochastically sampled masks on one layer at "
         "a time",
     )
+    ci_recon_subset_coeff: NonNegativeFloat | None = Field(
```
danbraunai, line 164: I'd change this to ci_masked_recon_subset_coeff now since we're going to have to do it at some stage.

spd/utils/component_utils.py
```diff
@@ -22,24 +22,87 @@ def _sample_stochastic_mask(
     return causal_importances + (1 - causal_importances) * rand_tensor
 
 
+RoutingType = Literal["uniform_k-stochastic", "all"]
```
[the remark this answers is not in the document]
danbraunai: My initial thought was to just make "routing" a bool. Though I guess it is very likely we'll want other options. Unsure whether we should just make it bool and change it later if we have other types or to leave it as is.

If we made it bool I'd just put the description in the one function that this type is used rather than globally.

spd/utils/component_utils.py
```diff
@@ -22,24 +22,87 @@ def _sample_stochastic_mask(
...
+"""
+
+
+def rand_perm(shape: tuple[int, ...], dim: int, device: torch.device | str) -> Int[Tensor, "... k"]:
+    noise = torch.rand(shape, device=device)
+    # turn values into ranks via double argsort trick. (for example: [0.8, 0.2, 0.3] -> [2, 0, 1])
+    return noise.argsort(dim=dim).argsort(dim=dim)
+
+
+def sample_uniform_k_subset_routing_masks(
```
danbraunai, line 41: I think this guy could do with a couple of unittests. Just some that show that we get samples in the range that we'd expect for (batch,) and (batch, seq).

spd/utils/component_utils.py
```diff
@@ -22,24 +22,87 @@ def _sample_stochastic_mask(
...
+    for each position, sample k from [1, n_modules], then route to components for k out of `n_modules` modules
+all:
+    use components for all positions
+"""
+
+
+def rand_perm(shape: tuple[int, ...], dim: int, device: torch.device | str) -> Int[Tensor, "... k"]:
+    noise = torch.rand(shape, device=device)
+    # turn values into ranks via double argsort trick. (for example: [0.8, 0.2, 0.3] -> [2, 0, 1])
+    return noise.argsort(dim=dim).argsort(dim=dim)
```
danbraunai, line 38: I'd also either prefix an underscore (`_rand_perm`) or if keeping it public I'd add a couple of unittests

danbraunai (approved):
Looks good. Before merging yourself:
1. Address minor comments
2. Confirm with Lucius/Lee about the algorithm

I made some edits to the description to hopefully make it a bit clearer.

spd/utils/component_utils.py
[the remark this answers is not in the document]
oli-clive-griffin: I think "routing" is a sufficiently nuanced concept that a descriptive literal makes sense here. I'd also defer to [this blog](https://ntietz.com/blog/that-boolean-should-probably-be-something-else/) for a general argument in favour of enums/literals instead of bools almost everywhere


like, even now, it's not obvious that routing: bool means "should we use that specific uniform-k subset routing". The right boolean field name right now would be something like "uniform_k_stochastic_routing: bool" which is clunky imo.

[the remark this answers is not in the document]
danbraunai: Yeah fair enough. On board with this here.

Thanks for sharing. I do think readability is aided quite a bit by a clear bool when you're sure the type won't change though. I suppose it's pretty rare when you have that confidence.

PR #172, opened 2025-09-29
Oli clustering refactor

mivanit:
taking a look at @oli-clive-griffin's changes to `feature/clustering`


spd/clustering/s4_compute_distances.py
mivanit, line 1: noting here that `create_clustering_report` is now just called in main instead of in `s4_compute_distances`

mivanit:
ok claude hallucinated pretty much everything in that report. def of `merge_iteration` definitely has type hints, and type checks are passing.

PR #180, opened 2025-10-03
Hidden act recon loss

danbraunai:
## Description
- Implementing #158 in the new loss/eval setup
- Adds it as an eval in all the default configs.

Notes:
1. Currently, it's implemented as a separate metric from stochastic recon. These could technically share the same component forward pass to speed things up. But it's much nicer in the code having separation between metrics. If we find ourselves using this as a loss function frequently, we can optimise it
2. It's only applied to the regular stochastic forward pass, and not forward passes which use routing (like those used for stochastic_recon_subset_loss

## Motivation and Context
Copied from #158:
> Main motivation: We might be able to identify pathological runs earlier if we are tracking hidden activation reconstruction. We might have more quickly noticed SPD using the last layer to cheat, for instance.
Secondary motivation: We might get some benefits for weakly optimizing for hidden layer activation reconstruction loss as an auxiliary loss. It might make it harder for networks to cheat in general, for instance. But it may also reconstruct unwanted noise, so we wouldn't want to optimize for it too strongly.

## How Has This Been Tested?
- Added test_stochstic_hidden_acts_recon which manually calculates the hidden act MSE for a small dummy model and compares to the new function.
- Here's a [run](https://wandb.ai/goodfire/spd/runs/77plxnev) of ss_llama. hidden_act_recon goes down at least.

No


danbraunai:
RFR. @leesharkey probably best if you review this one (would be a good way to check out how the new metric structure works and also verify that the logic is correct).

oli-clive-griffin (approved):
Beautiful, this is such a cool example of the refactor working super well imo

spd/metrics/stochastic_hidden_acts_recon.py
```diff
@@ -0,0 +1,129 @@
...
+        n_mask_samples,
+        batch,
+        pre_weight_acts,
+        ci,
+        weight_deltas,
+    )
+    return _stochastic_hidden_acts_recon_compute(sum_mse, n_examples)
+
+
+class StochasticHiddenActsRecon(Metric):
```
leesharkey, line 84: Small nit: Some of the metric classnames are a big confusing, in that it's not clear what they are. I think this comes from not having specified whether they're a loss or a metric. If everything has 'Metric' in the class name (rather than some having 'Loss' and others having nothing, then I think it might be clearer to understand what they are when reading. Not a requirement to pass the review, more a question about whether or not you agree it'd be better to have 'Metric' in the classname? I'd lean yes but not strongly opinionated on it.

leesharkey (approved):
Logic makes sense. Looks good to me. Keen to get this in. 

I do have some concerns about having to run a whole separate forward pass for this. I think it's quite likely that we'll want to use this in training in future, but that's a personal take for now. I'll be interested to see how much of a slowdown it is to run the forward pass twice. 

Feel free to merge when ready.

spd/metrics/stochastic_hidden_acts_recon.py
danbraunai: Oh shoot. I meant for this to be `StochasticHiddenActsReconLoss`. The current way I was doing it was that everything with Loss in the name can be used as a loss function (and thus has the functional form rather than just the class form. I just forgot to add Loss to this one. I've added it now to this one.

I prefer this to labelling everything with metric. Thoughts?

leesharkey:
Unsure if this is intended behavior, but the following diff was required to get the hidden acts recon loss working on a branch that I'd merged main into (`lee/res_plots`, but shouldn't be relevant):

```
(spd) lee@tenant-ac-goodfre-h100-dev-194-031:~/spd lee/res_plots$ git diff
diff --git a/spd/configs.py b/spd/configs.py
index 3776b493..496a32f5 100644
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -111,7 +111,11 @@ class StochasticReconSubsetCEAndKLConfig(BaseModel):
     exclude_patterns: dict[str, list[str]] | None
 
 
-class StochasticHiddenActsReconLossConfig(BaseModel):
+class StochasticHiddenActsReconLossConfig(TrainMetricConfig):
+    classname: Literal["StochasticHiddenActsReconLoss"] = "StochasticHiddenActsReconLoss"
+
+
+class StochasticHiddenActsReconLossEvalConfig(BaseModel):
     classname: Literal["StochasticHiddenActsReconLoss"] = "StochasticHiddenActsReconLoss"
 
 
@@ -130,7 +134,7 @@ TrainMetricConfigType = (
     | StochasticReconLayerwiseLossTrainConfig
     | StochasticReconLossTrainConfig
     | StochasticReconSubsetLossTrainConfig
-    | StochasticHiddenActsReconConfig
+    | StochasticHiddenActsReconLossConfig
 )
 EvalMetricConfigType = (
     CEandKLLossesConfig
@@ -142,7 +146,7 @@ EvalMetricConfigType = (
     | PermutedCIPlotsConfig
     | UVPlotsConfig
     | StochasticReconSubsetCEAndKLConfig
-    | StochasticHiddenActsReconLossConfig
+    | StochasticHiddenActsReconLossEvalConfig
 )
 MetricConfigType = TrainMetricConfigType | EvalMetricConfigType
 
diff --git a/spd/losses.py b/spd/losses.py
index d4cc80fd..f4b83571 100644
--- a/spd/losses.py
+++ b/spd/losses.py
@@ -10,7 +10,7 @@ from spd.configs import (
     CIMaskedReconSubsetLossTrainConfig,
     FaithfulnessLossTrainConfig,
     ImportanceMinimalityLossTrainConfig,
-    StochasticHiddenActsReconConfig,
+    StochasticHiddenActsReconLossConfig,
     StochasticReconLayerwiseLossTrainConfig,
     StochasticReconLossTrainConfig,
     StochasticReconSubsetLossTrainConfig,
@@ -22,7 +22,7 @@ from spd.metrics import (
     ci_masked_recon_subset_loss,
     faithfulness_loss,
     importance_minimality_loss,
-    stochastic_hidden_acts_recon,
+    stochastic_hidden_acts_recon_loss,
     stochastic_recon_layerwise_loss,
     stochastic_recon_loss,
     stochastic_recon_subset_loss,
@@ -126,12 +126,12 @@ def compute_total_loss(
                     ci=ci,
                     weight_deltas=weight_deltas,
                 )
-            case StochasticHiddenActsReconConfig():
+            case StochasticHiddenActsReconLossConfig():
                 if pre_weight_acts is None:
                     raise ValueError(
-                        "pre_weight_acts is required for StochasticHiddenActsRecon but was not provided"
+                        "pre_weight_acts is required for StochasticHiddenActsReconLoss but was not provided"
                     )
-                loss = stochastic_hidden_acts_recon(
+                loss = stochastic_hidden_acts_recon_loss(
                     model=model,
                     sampling=sampling,
                     use_delta_component=use_delta_component,
```

PR #183, opened 2025-10-06
Use pre-built MPI docker image in CI

### LINKED ISSUE #110: Add custom docker image with mpi for CI
Currently, installing openmpi takes >1 minute in every workflow. It'd be nice if our CI ran on an image which had openmpi preinstalled. There are some existing public ones like https://hub.docker.com/r/mfisherman/openmpi, but they look more complex than I'd like. We just want ubuntu-latest + openmpi.

If we go with a public one, we should fix a specific version rather than use "latest" for security reasons. Otherwise we can of course build our own.

danbraunai:
## Description
Reduces CI runtime by >2 mins by loading a custom docker image instead of the default from ubuntu-latest

- Adds a Dockerfile that installs MPI (and git).
- Gives an option to build and publish this dockerfile to github's **public** container registry in a new CI job.
- **This job must be run manually whenever we want to build the image again.**

## Related Issue
Closes #110 

## Motivation and Context
CI currently has to install openmpi everytime. This takes >2 minutes.

## How Has This Been Tested?
See the CI runs in this PR.

No

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:latest
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y -q \
+    openmpi-bin \
+    libopenmpi-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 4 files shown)


danbraunai:
RFR. @oli-clive-griffin you able to look at this one?

Issue #87, opened 2025-07-28
Fix loss spikes


danbraunai:
In the run on all MLP layers of SS 1.25M ([here](https://wandb.ai/goodfire/spd/reports/Simple-Stories-1-25M-MLPs--VmlldzoxMzc2MjAyMQ)), we're seeing a lot of loss spikes. We should see if reducing the learning rate fixes this. Otherwise we might want to look into initilization (I think @oli-clive-griffin had some unmerged tweaks here) and/or optimizer strategies that are known to stabilize Adam during training.

danbraunai:
A smaller learning rate at least reduces the spikes https://wandb.ai/goodfire/spd/runs/37lwt4fr. We can possibly remove them entirely with an even smaller learning rate.

Though note that it would be desirable if we could come up with another fix that allows for higher learning rates. This would be important but lower priority if we do manage to fix the loss spikes with a smaller lr.

PR #187, opened 2025-10-06
allow installing cpu-only torch in CI

mivanit:
## Motivation and Context

CI runs were failing due to insufficient disk space:
https://github.com/goodfire-ai/spd/actions/runs/18292296525/job/52082212638

a few GB of space are taken up by the docker container, but the main culprits are:
- UV installs torch with cuda deps, despite no cuda device in CI
- UV installs by copy since symlinks/hardlinks seem to fail, which leaves a giant cache that is not being used

## Description
- add (conflicting) extras `cpu` and `cu128` with their respective indecies
  - `make install` and `make install-dev` install with `cu128` by default, configurable via `CUDA_EXTRA=foo make install-dev`
    - unfortunately `uv sync` as of `uv==0.8.23` will now install torch with cpu by default, no way to change this as far as I can tell. see: https://docs.astral.sh/uv/guides/integration/pytorch/#automatic-backend-selection
- modified installation of deps in CI:
  - install via `uv sync --extra cpu --link-mode copy`
  - install followed by `uv cache clean`

This results in a huge increase in disk space and a modest reduction in CI installation time:

|        |  disk space  | install time         | link  |
| :---:  | :---:        | :---:                | :---: |
| Before | 6.7GB        | ~60s                 | [`actions/runs/18293012119/job/52084684178`](https://github.com/goodfire-ai/spd/actions/runs/18293012119/job/52084684178)
| After  | 21GB         | 12s + 2s cache clean | [`actions/runs/18294570560/job/52089849831`](https://github.com/goodfire-ai/spd/actions/runs/18294570560/job/52089849831) |

> Note: "disk space" meaning "disk space available right before tests start"

## Related Issue

CI runs were failing due to insufficient disk space:

https://github.com/goodfire-ai/spd/actions/runs/18292296525/job/52082212638

possibly, if any of our tests rely on torch+cuda being installed

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -32,7 +32,10 @@ jobs:
         run: uv python install 3.12
 
       - name: Install dependencies
-        run: uv sync
+        run: uv sync --extra cpu --link-mode copy
+
+      - name: clear uv cache
+        run: uv cache clean || true
 
       - name: Print dependencies
         run: uv pip list
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,15 @@
 # setup
+CUDA_EXTRA ?= cu128
+
 .PHONY: install
 install: copy-templates
-	uv sync --no-dev
+	uv sync --no-dev --extra ${CUDA_EXTRA}
 
 .PHONY: install-dev
 install-dev: copy-templates
-	uv sync
+	uv sync --extra ${CUDA_EXTRA}
 	pre-commit install
+
 
 .PHONY: copy-templates
 copy-templates:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ urls = { "Homepage" = "https://github.com/goodfire-ai/spd" }
 license = { text = "MIT" }
 readme = "README.md"
 dependencies = [
+    # NOTE: if you need to change either torch dep, also change them in [project.optional-dependencies]
     "torch>=2.6",
     "torchvision",
     "pydantic",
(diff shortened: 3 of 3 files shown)


mivanit:
added the pin of `transformers==4.53.2`, initially done in #43 , which mitigates but does not resolve #87

mivanit:
also note that i rebuilt the docker container with `make` installed: https://github.com/goodfire-ai/spd/actions/runs/18315313364/job/52153891071

spd/utils/conv1d.py
```diff
@@ -0,0 +1,5 @@
+"""wrapper for Conv1D, see https://github.com/goodfire-ai/spd/issues/139"""
+
+from transformers.pytorch_utils import Conv1D as RadfordConv1D
+
+__all__ = ["RadfordConv1D"]
```
danbraunai, line 5: I'd remove this. We're not using `__all__` in other parts of the codebase. I think `from X import *` shouldn't be encouraged. the `__all__` does also affect "re-exporting", but I think this should be handled with using "as" everywhere. See `spd/metrics/__init__.py` for what I'm talking about.

mivanit: just ended up getting rid of this file [70facfb](https://github.com/goodfire-ai/spd/pull/187/commits/70facfb047bc2b00f4c7105d0379837c40246099)

danbraunai: hah yeah I've been in some fights with ruff before. If you did have to keep this file for whatever reason (I see you removed it so you don't), I'd probably just add a ruff ignore to the file/line.

PR #189, opened 2025-10-07
Fix metric structure and hidden_act_recon

danbraunai:
## Description
- Change the type of `TrainMetricConfig.coeff` from `float` to `float | None`. When a metric that inherits from this class is specified in loss_metric_configs, we assert that coeff is not None. When it's specified in eval_metric_configs, we don't care.
- Make hidden act recon a "loss metric" (i.e. make it inherit from TrainMetricConfig, move it from EvalMetricConfigType to TrainMetricConfigType, handle its initialization in losses.py).
- Rename EvalMetricConfigType to EvalOnlyMetricConfigType, to make it clear than nothing there can also be a loss metric (if it was, it'd belong in TrainMetricConfigType.

## Motivation and Context
We had a couple of bugs in #162:
1. hidden_act_recon was an "eval" and not a "train metric". Concretely, it didn't inherit from TrainMetricConfig. This means that it couldn't be used as a train metric. Lee noticed this [here](https://github.com/goodfire-ai/spd/pull/180#issuecomment-3376534046).
2. There was no way to specify a loss metric in eval_metric_configs. This means that we couldn't eval a loss metric unless we also wanted to use it as a loss.

## How Has This Been Tested?
Manual running with StochasticHiddenActsReconLoss as in loss_metric_configs and in eval_metric_configs (and in both).

No

### DIFF
--- a/spd/configs.py
+++ b/spd/configs.py
@@ -23,11 +23,11 @@ from spd.spd_types import ModelPath, Probability
 from spd.utils.general_utils import BaseModel
 
 
-#### Train Metric Configs ####
+#### Metrics that can be used in training (or eval) ####
 class TrainMetricConfig(BaseModel):
-    coeff: float = Field(
-        ...,
-        description="Coefficient used for weighting into loss/total.",
+    coeff: float | None = Field(
+        default=None,
+        description="Loss coefficient. Used when metric is in loss_metric_configs.",
     )
 
 
@@ -68,7 +68,11 @@ class StochasticReconSubsetLossTrainConfig(TrainMetricConfig):
     classname: Literal["StochasticReconSubsetLoss"] = "StochasticReconSubsetLoss"
 
 
-#### Eval Metric Configs ####
+class StochasticHiddenActsReconLossConfig(TrainMetricConfig):
+    classname: Literal["StochasticHiddenActsReconLoss"] = "StochasticHiddenActsReconLoss"
+
+
+#### Metrics that can only be used in eval ####
 class CEandKLLossesConfig(BaseModel):
     classname: Literal["CEandKLLosses"] = "CEandKLLosses"
     rounding_threshold: float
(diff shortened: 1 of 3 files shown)


oli-clive-griffin:
Little tired but so take with a grain of salt but:  A clean way to [Make Invalid States Unrepresentable](https://www.youtube.com/watch?v=z-0-bbc80JM) here would be something like:

```
loss_metric_configs: list[{coeff: float, metric: Metric}]
eval_metric: Metric
```
No need for runtime type checking that way

danbraunai:
@oli-clive-griffin 
> Little tired but so take with a grain of salt but: A clean way to [Make Invalid States Unrepresentable](https://www.youtube.com/watch?v=z-0-bbc80JM) here would be something like:
> 
> ```
> loss_metric_configs: list[{coeff: float, metric: Metric}]
> eval_metric: Metric
> ```
> 
> No need for runtime type checking that way

This is a good idea. [Here's](https://github.com/goodfire-ai/spd/pull/194) an implementation of it.

But I don't feel inclined to implement it because:
1. It adds more nesting to the config:
```
loss_metric_configs:
  - coeff: 0.1
    metric:
      classname: "ImportanceMinimalityLoss"
      pnorm: 2.0
      p_anneal_start_frac: 0.0
      p_anneal_final_p: 0.7
      p_anneal_end_frac: 1.0
```
This is probably a little more difficult for users to understand (unsure about this).
2. It adds some complexity to all the new utilities in this PR for handling sweeps and wandb run names.
3. (I think this is the biggest issue) It makes the sweep configs weird. We can still use the structure:
```
  global:
    seed:
      values: [0, 1]
    loss_metric_configs:
      - classname: "ImportanceMinimalityLoss"
        coeff:
          values: [0.1, 0.2]
        pnorm:
          values: [0.9, 1.0]
```
and then just handle this in the backend when building what we're sweeping over, but a user is going to be pretty confused here with this difference in structure between a regular config and a sweep config. I think other options of sweep structure are going to be pretty difficult/confusing.

So I'm inclined to leave it as is, despite agreeing that it's nice to avoid having invalid states being representable. Thoughts?

PR #219, opened 2025-10-16
Handle Metric config overlaps

danbraunai:
## Description
- If both loss and eval metric configs exist, only run evals on the version specified in the eval_metric_configs.
- Throws a warning if there is a duplicate
- Renamed TrainMetric to LossMetric. This is separate to the main PR but I threw it in. Doesn't affect any user-facing configs, and I found the Train vs Loss metric naming very confusing (we used both arbitrarily).

## Motivation and Context
We sometimes want to use different configs for eval than what we use for training (e.g. n_steps in PGD loss, which isn't yet in main).

## How Has This Been Tested?
Manually looked at it. All good.

No

### DIFF
--- a/spd/eval.py
+++ b/spd/eval.py
@@ -226,7 +226,7 @@ def init_metric(
 
 
 def evaluate(
-    metric_configs: list[MetricConfigType],
+    eval_metric_configs: list[MetricConfigType],
     model: ComponentModel,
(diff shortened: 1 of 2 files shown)


tests/metrics/test_alive_components_distributed.py
danbraunai: I dunno man, ruff wouldn't let me commit without this.

oli-clive-griffin (approved):

PR #222, opened 2025-10-17
Add PGD metrics

oclivegriffin:
## Description
- Adds pgd losses (PGDReconLoss, PGDReconSubsetLoss, PGDReconLayerwiseLoss). As with all of our metrics, these can be used in train or eval.
- The PGD optimization is synced between processes, and is thus independent of the number of distributed processes. I.e. if you use --dp 4, then each of the 4 processes will have the same adversarial sources at each step of the PGD optimization process.

## Motivation and Context
We want to be able to show whether we can adversarially set masks to components that have a CI value of <1, and not break our decomposition.

## How Has This Been Tested?
- [Report](https://wandb.ai/goodfire/spd/reports/Robustness-of-PGDReconLoss-with-different-topologies--VmlldzoxNDg1MTgyMA) showing that training with different topologies (e.g. dp2, dp8, no data parallel), gives similar or the same results. Complete determinism is tricky. Sometimes we do get identical results (e.g. between the nodp and dp8-no-asserts run in the report), but other times I think we get some floating point errors accumulating in the actual training, which causes some different results in the PGD evals.

No

### DIFF
--- a/spd/metrics/pgd_utils.py
+++ b/spd/metrics/pgd_utils.py
@@ -121,6 +121,16 @@ def _interpolate_component_mask(
     ci: dict[str, Float[Tensor, "*batch_dims C"]],
     adv_sources_components: Float[Tensor, "n_layers *batch_dims C"],
 ) -> dict[str, Float[Tensor, "*batch_dims C"]]:
+    """Set the mask value to ci + (1 - ci) * adv_sources_components.
+
+    NOTE: This is not ideal. Suppose ci is 0.2 and adv_sources_components is 0.8. Then the mask
+    value will be 0.2 + (1 - 0.2) * 0.8 = 0.84. It would make more sense to set the mask value to
+    0.8 directly, since this is the value output by PGD. If we wanted to instead use the more
+    natural setup of mask = max(ci, adv_sources_components), we would need to work out how to
+    maintain the gradient flow. We feel that this likely isn't a big problem as it is right now
+    since the change would just give PGD more optimization power, and we already get a very bad
+    loss value for it.
+    """
     assert torch.all(adv_sources_components <= 1.0) and torch.all(adv_sources_components >= 0.0)
     assert adv_sources_components.shape[0] == len(ci)
     assert all(ci[k].shape[-1] == adv_sources_components.shape[-1] for k in ci)


spd/experiments/lm/ss_llama_config.yaml
```diff
@@ -26,16 +26,32 @@ identity_module_patterns: null
...
+    init: "random"
+    step_size: 0.01
+    n_steps: 3
+    mask_scope: "shared_across_batch"
+    coeff:
+      type: "cosine"
+      start_value: 0.0
+      end_value: 0.0
+      start_frac: 0.0
+      end_frac: 1.0
```
danbraunai: Best remove from the config (the coeffs are 0 anyway). Can add it to the eval_loss_metrics instead

spd/utils/component_utils.py
```diff
@@ -96,11 +76,37 @@ def sample_uniform_k_subset_routing_masks(
...
+"""How to choose which (batch,) or (batch, seq_len) positions to route to components or target.
+
+uniform_k-stochastic:
+    for each position, sample k from [1, n_modules], then route to components for k out of
+    `n_modules` modules
+all:
+    use components for all positions
+given:
+    use the given routing masks
+"""
```
danbraunai: - given should be deleted

spd/utils/component_utils.py
```diff
@@ -96,11 +76,37 @@ def sample_uniform_k_subset_routing_masks(
...
+    routing: RoutingType,
+    leading_dims: tuple[int, ...],
+    module_names: list[str],
+    device: torch.device | str,
+) -> RoutingMasks:
+    match routing:
+        case "all":
+            return "all"
+        case "uniform_k-stochastic":
+            return sample_uniform_k_subset_routing_masks(leading_dims, module_names, device)
```
danbraunai: I think I'd prefer just not having this function and doing the if statement explicitly at call sites. Reasons:
1. The function does nothing except in one specific case (uniform_k-stochastic)
2. 3/4 arguments to the function are only used in that one case.
3. There are only 2 call sites to this.

spd/utils/general_utils.py
```diff
@@ -341,6 +341,24 @@ def get_obj_devices(d: CanGetDevice) -> set[torch.device]:
...
+    d1: dict[str, T1], d2: dict[str, T2], d3: dict[str, T3], /
+) -> dict[str, tuple[T1, T2, T3]]: ...
+
+
+def zip_dicts(*dicts: dict[str, Any]) -> dict[str, tuple[Any, ...]]:
+    all_keys = set(dicts[0])
+    for d in dicts[1:]:
+        assert set(d.keys()) == all_keys
+        all_keys.update(d.keys())
+    return {k: tuple(d[k] for d in dicts) for k in all_keys}
```
danbraunai: - The assert and then the update doesn't make sense. I think you want to get rid of the update call.

Fwiw I was wondering if you could get rid of the overloads. AI suggested:
```
from typing import Mapping, Tuple
from typing_extensions import TypeVarTuple, Unpack  # on 3.12+ you can import from typing

Ts = TypeVarTuple("Ts")

def zip_dicts(
    *dicts: Unpack[Tuple[Mapping[str, Unpack[Ts]]]]
) -> dict[str, Tuple[Unpack[Ts]]]:
    if not dicts:
        return {}
    first_keys = set(dicts[0])
    for d in dicts[1:]:
        assert set(d) == first_keys, "All dictionaries must have the same keys"
    return {k: tuple(d[k] for d in dicts) for k in dicts[0]}

```
Which I have not verified. Don't care that much about overloads if it's a random utility like this. But I would prefer something simpler here. If can't think of anything then dw

spd/configs.py
```diff
@@ -151,6 +202,8 @@ class UVPlotsConfig(BaseConfig):
 
 TaskConfig = TMSTaskConfig | ResidMLPTaskConfig | LMTaskConfig | IHTaskConfig
 
+SamplingType = Literal["continuous"] | Literal["binomial"]
```
danbraunai:
```suggestion
SamplingType = Literal["continuous", "binomial"]
```

pyproject.toml
```diff
@@ -28,6 +28,8 @@ dependencies = [
     # see:  https://github.com/huggingface/datasets/issues/6980  https://github.com/huggingface/datasets/pull/6991  (fixed in https://github.com/huggingface/datasets/releases/tag/2.21.0 )
     "datasets>=2.21.0",
     "simple_stories_train @ git+https://github.com/goodfire-ai/simple_stories_train.git@dev",
+    "fastapi",  # Latest compatible version
+    "uvicorn",
```
danbraunai: unrelated

uv.lock
danbraunai: unrelated changes in here related to the app too

spd/metrics/pgd_utils.py
```diff
@@ -0,0 +1,188 @@
...
+                    adversarial_component_sample_points, adversarial_weight_delta_masks
+                )
+            )
+
+            grads = torch.autograd.grad(
+                obj,
+                adv_vars,
+                retain_graph=False,
+                create_graph=False,
+                allow_unused=True,
```
danbraunai: I'd remove this line. It should be the default, which is False. We don't want this to succeed if elements in adv_vars is not in the computational graph

spd/metrics/pgd_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        adv_vars.extend(list(adversarial_weight_delta_masks.values()))
+
+    for _ in range(int(pgd_config.n_steps)):
+        assert all(v.requires_grad for v in adv_vars)
+        assert all(v.grad is None for v in adv_vars)
+
+        with torch.enable_grad():
+            obj = objective_fn(
+                *maybe_repeat_to_batch_shape(
+                    adversarial_component_sample_points, adversarial_weight_delta_masks
```
danbraunai: I'd be tempted to use "adv" instead of "adversarial" throughout this function. I think it would reduce the space in places like this.

spd/metrics/pgd_utils.py
```diff
@@ -0,0 +1,188 @@
...
+) -> dict[str, Float[Tensor, "... C"]]:
+    component_mask = {}
+    for module_name in ci:
+        assert torch.all(component_sample_points[module_name] <= 1.0)
+        scaled_noise_to_add = (1 - ci[module_name]) * component_sample_points[module_name]
+        assert torch.all(scaled_noise_to_add >= 0.0)
+        assert torch.all(scaled_noise_to_add <= 1.0 - ci[module_name])
+        component_mask[module_name] = (
+            ci[module_name] + (1 - ci[module_name]) * component_sample_points[module_name]
+        )
```
danbraunai: Duplicated computation here

spd/losses.py
```diff
@@ -22,12 +28,29 @@
...
     stochastic_hidden_acts_recon_loss,
     stochastic_recon_layerwise_loss,
     stochastic_recon_loss,
     stochastic_recon_subset_loss,
 )
 from spd.models.component_model import CIOutputs, ComponentModel
+from spd.scheduling import get_cosine_schedule_value, get_linear_schedule_value
+
+
+def get_loss_coeff(
```
danbraunai: maybe just call this get_coeff, as we might use it for things that aren't loss coefficients in the future.

spd/losses.py
```diff
@@ -22,12 +28,29 @@
...
+    coeff: LinearSchedule | CosineSchedule | float | int,
+    current_frac_of_training: float,
+) -> float:
+    match coeff:
```
danbraunai:
```suggestion
) -> float:
    """Get the coefficient for the current step of training."""
    match coeff:
```

spd/run_spd.py
```diff
@@ -295,7 +294,7 @@ def optimize(
 
         # --- Evaluation --- #
         if step % config.eval_freq == 0:
-            with torch.inference_mode():
+            with torch.no_grad():
                 slow_step: bool = (
                     config.slow_eval_on_first_step
```
danbraunai: Any reason for this? I don't think this case brought up by ptrblck appolies: https://discuss.pytorch.org/t/pytorch-torch-no-grad-vs-torch-inference-mode/134099?u=timgianitsos

spd/utils/general_utils.py
oli-clive-griffin: oh yea if that typing works then it's way nicer!

oli-clive-griffin: And yea this impl is shit. I'm basically trying to say: "assert all dicts have the same keys", can remove the update line

spd/metrics/pgd_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    def maybe_repeat_to_batch_shape(
+        component_sample_points: dict[str, Float[Tensor, "... C"]],
+        weight_delta_mask: dict[str, Float[Tensor, "..."]] | None,
+    ) -> tuple[dict[str, Float[Tensor, "... C"]], dict[str, Float[Tensor, "..."]] | None]:
+        return (
+            {k: v.expand(*batch_dims, -1) for k, v in component_sample_points.items()},
+            {k: v.expand(*batch_dims) for k, v in weight_delta_mask.items()}
+            if weight_delta_mask is not None
+            else None,
+        )
```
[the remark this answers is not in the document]
oli-clive-griffin: yea definitely don't love how this is handled at all.

A nicer but less safe way would be to just allow passing in masks with singleton dims, and broadcasting inside the component hook. There's currently a shape assertion that means this would fail, but perhaps the best thing is to just remove that and do the singleton strategy.

I initially though it'd be safer to handle it all at this level but it's become so gross now that we should probably switch course.

spd/metrics/pgd_utils.py
oli-clive-griffin: completely agree, this was just a holdover from the original vibe-implementation. My bad. definitely agree with the fail-early sentiment too btw

oli-clive-griffin:
oh also - I can't approve bc I'm author, but pending the cosine fix, then: ✅

tests/test_scheduling.py
```diff
@@ -0,0 +1,188 @@
...
+        assert math.isclose(result_at_start_frac, expected, rel_tol=1e-9)
+
+        # Just before end_frac (since at end_frac it returns end_value directly)
+        result_just_before_end = get_cosine_schedule_value(schedule, current_frac_of_training=0.74)
+        expected = 0.0 + 0.5 * (1.0 - 0.0) * (1 + math.cos(math.pi * 0.74))
+        assert math.isclose(result_just_before_end, expected, rel_tol=1e-9)
+
+        # At end_frac, should return end_value directly
+        result_at_end_frac = get_cosine_schedule_value(schedule, current_frac_of_training=0.75)
+        assert result_at_end_frac == 0.0
```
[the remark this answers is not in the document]
danbraunai: See comment [here](https://github.com/goodfire-ai/spd/pull/222#issuecomment-3455489663).

PR #230, opened 2025-10-21
Add ss_llama_simple in place of ss_llama

danbraunai:
## Description
**Everyone should use the new ss_llama_simple instead of ss_llama. You will need to `uv sync` after pulling this.**

Created a new ss_llama_simple model in https://github.com/goodfire-ai/simple_stories_train/tree/dev. This is the same as the regular llama but:
- We don't use flash attention
- We train in float32 instead of (mostly) bfloat16. We needed to do this to avoid divergence
- We split up kv_attn into two matrices
- We change the paths of all modules so they are similar to gpt2. The paths are now:
```
  [
    "h.*.mlp.gate_proj",
    "h.*.mlp.up_proj",
    "h.*.mlp.down_proj",
    "h.*.attn.q_proj",
    "h.*.attn.k_proj",
    "h.*.attn.v_proj",
    "h.*.attn.o_proj",
  ]
```

Removed ss_llama_config.yaml.

## Motivation and Context
- ss_llama was trained with simplestories, but was uploaded to huggingface. The model we decompose is thus the huggingface version.
- This makes it quite awkward to interpret, because the huggingface model is complex and has a lot of options.

## How Has This Been Tested?
Testing that ss_llama gives the same as ss_llama_simple with a couple of different configs:
1 (Lucius asked me for this one).
- ~SPD [Run](https://wandb.ai/goodfire/spd/runs/qwbm917v) of ss_llama_simple (This run accidentally didn't use the sigmoid fix)~
- SPD [Run](https://wandb.ai/goodfire/spd/runs/hhekh9wt) of ss_llama_simple with batch size 8 (accidental). It still gets similar enough results to the below
- SPD [Run](https://wandb.ai/goodfire/spd/runs/3rthvni3) of ss_llama

2.
- SPD [Run](https://wandb.ai/goodfire/spd/runs/g19cvhed) of ss_llama
- SPD [Run](https://wandb.ai/goodfire/spd/runs/va94mjf5) of ss_llama_simple. This is currently IMO close enough to the run above. I expect some differences due to randomness and because it was trained on a different tokenizer (4019 tokens compared to 4096).
- Also, we verified that the pretrained models had similar val losses

Relatedly, this set off a full investigation of running SPD on models trained with a different number of steps. Wandb report [here](https://wandb.ai/goodfire/spd/reports/Effect-of-target-model-training-steps-on-SPD--VmlldzoxNDgwMTEwMQ).

No

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1078,7 +1078,7 @@ name = "nvidia-cudnn-cu12"
 version = "9.10.2.21"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "nvidia-cublas-cu12" },
+    { name = "nvidia-cublas-cu12", marker = "sys_platform == 'linux'" },
 ]
 wheels = [
     { url = "https://files.pythonhosted.org/packages/ba/51/e123d997aa098c61d029f76663dedbfb9bc8dcf8c60cbd6adbe42f76d049/nvidia_cudnn_cu12-9.10.2.21-py3-none-manylinux_2_27_x86_64.whl", hash = "sha256:949452be657fa16687d0930933f032835951ef0892b37d2d53824d1a84dc97a8", size = 706758467, upload-time = "2025-06-06T21:54:08.597Z" },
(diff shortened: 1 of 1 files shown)


oli-clive-griffin (approved):
LGTM!

Issue #221, opened 2025-10-17
Pydantic UnsupportedFieldAttributeWarning: warning


danbraunai:
We get this when running anything now:
```
/mnt/polished-lake/home/braun/spd-main/.venv/lib/python3.13/site-packages/pydantic/_internal/_generate_schema.py:2249: UnsupportedFieldAttributeWarning: The 'repr' attribute with value False was provided to the `Field()` function, which has no effect in the context it was used. 'repr' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
/mnt/polished-lake/home/braun/spd-main/.venv/lib/python3.13/site-packages/pydantic/_internal/_generate_schema.py:2249: UnsupportedFieldAttributeWarning: The 'frozen' attribute with value True was provided to the `Field()` function, which has no effect in the context it was used. 'frozen' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
```

It's due to https://github.com/wandb/wandb/pull/10651 and should be fixed at the next wandb release, whenever that is. I'm inclined to just ignore it until then, but if someone really wanted they could pin pydantic<2.12 and then release it when wandb releases a new fix.

oli-clive-griffin:
ahhh nice. out of interest how did you find this out?

danbraunai:
Searched the error in google and the github issue on wandb came up

oli-clive-griffin:
🫠

PR #232, opened 2025-10-23
Fix dependency issues

oclivegriffin:
## Description
CI is currently broken on main ([example failing build](https://github.com/goodfire-ai/spd/actions/runs/18716967662/job/53378759193)), thanks to @mivanit for suggesting the fix to pin basedpyright.

Figured it'd also be a good time to fix [this](https://github.com/goodfire-ai/spd/issues/221)

## Related Issue
closes https://github.com/goodfire-ai/spd/issues/221

## How Has This Been Tested?
N/A

nope

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 dependencies = [
     "torch>=2.6",
     "torchvision>=0.23,<0.24",
-    "pydantic<2.12",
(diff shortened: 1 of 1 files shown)


mivanit (changes requested):
IMO whenever pinning versions, especially when those changes set a max and not a min version, we should leave comments explaining why or linking to explanations.

danbraunai:
@oli-clive-griffin Please make an issue noting that the pins are due to this issue. This will remind us to unpin when it's fixed.

You can merge.

PR #239, opened 2025-10-28
Add multi-batch pgd metric

danbraunai:
[Updated by @oli-clive-griffin 10 Nov 2025]
see comment below for update

## Description
- Adds PGDGlobalReconLossConfig and PGDGlobalReconSubsetLossConfig. These can be added to eval_metric_configs and loss_metric_configs.
- They will calculate global PGD in `calc_global_pgd_metrics`, which is separate from the main `evaluate` function.
- Does not support layerwise. Would be a bit annoying to add so just avoided it for now. Can add in future if we want it.

## Motivation and Context
We'd like to have our PGD "shared_across_batch" be applied to much more than a single batch. E.g. we might want to create a single adversarial source per component for the full training dataset.

No

### DIFF
--- a/spd/scripts/run.py
+++ b/spd/scripts/run.py
@@ -163,7 +163,6 @@ def generate_commands(
             command = (
                 "NCCL_DEBUG=WARN "
                 "TORCH_NCCL_ASYNC_ERROR_HANDLING=1 "
-                "NCCL_ASYNC_ERROR_HANDLING=1 "
                 f"{mpi_prefix}"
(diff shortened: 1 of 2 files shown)


oli-clive-griffin:
@claude Can you please review this. The things I'm most worried about are:
- correct handling of gradients, how these propagate, whether they're detached and managed correctly, whether we've handled the distinction between training and eval modes correctly, whether we've correctly isolated pgd opimisation from the broader training loop etc.
- logical errors in the multi batch PGD setup

PR #245, opened 2025-11-04
Move to new cluster

oclivegriffin:
## Description
Adjusts everything necessary to work on the new H200 cluster.

## Related Issue
N/A

## How Has This Been Tested?
manually

No

### DIFF
--- a/tests/scripts_run/test_main.py
+++ b/tests/scripts_run/test_main.py
@@ -146,8 +146,9 @@ class TestSPDRun:
 
             # Should be a list of arguments
             assert isinstance(args, list)
-            assert args[0] == "python"
-            assert "_decomposition.py" in args[1]
+            assert args[0] == "NCCL_DEBUG=WARN"
+            assert args[1] == "python"
+            assert "_decomposition.py" in args[2]
 
             # Check for required arguments in the command
             cmd_str = " ".join(args)


spd/scripts/run.py
```diff
@@ -188,7 +188,7 @@ def generate_commands(
 
                 mpi_prefix = _build_mpi_prefix(run_id, cmd_idx, dp) if dp > 1 else ""
                 command = (
-                    f"{mpi_prefix}python {exp_config.decomp_script} --config_json '{config_json}' "
+                    f"{mpi_prefix} python {exp_config.decomp_script} --config_json '{config_json}' "
                     f"--sweep_id {run_id} "
```
mivanit: IMO this string processing way of handling commands is somewhat bug prone, but the `Command` class refactor I tried was overcomplicated. probably fine as long as we are careful

tests/test_distributed.py
```diff
@@ -161,6 +161,10 @@ def _run_experiment(
 
         cmd = [
             "mpirun",
+            "--bind-to",
+            "none",
+            "--map-by",
+            "slot",
```
mivanit: maybe we keep `EXTRA_MPI_ARGS` somewhere separate?

oli-clive-griffin:
> maybe good to centralize these options in some way to make things more maintainable?

Yep, good point, have abstracted out the partition name and just removed the mpi additions:

> explanations of the new mpirun opts would be good

I think I'm just going to remove these for now, merge and do some more testing, as I haven't been able to recreate the issue this supposedly solves

PR #248, opened 2025-11-11
run registry

mivanit:
## Description
structured way of keeping track of runs, instead of just pinning things on slack

## Related Issue
https://opensourcemechanistic.slack.com/archives/C08N7E5KNG7/p1762432153667829

## Questions

- what does the interface for interacting with this table look like, if we have one at all?
  - the runs are not "keyed", its just a list -- no way to get a specific element besides index. 
- format?
  - TOML would be preferable to YAML here IMO, but we are using YAML in most places already
- integration with wandb persistence?
  - we have a "keep" tag on wandb and periodically clear all runs older than X with that tag. perhaps tests for this should be "assert each referenced run has the `keep` tag"
- adding new runs
  - might be also good to have a persistent issue on the repo for adding new runs -- commenting a link is less friction than making a PR, pretty easy to periodically ask claude to turn all comments into a single PR

No.

### DIFF
--- a/spd/registry.py
+++ b/spd/registry.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Any
 
 import yaml
+from pydantic import BaseModel
 
 from spd.settings import REPO_ROOT
 from spd.spd_types import TaskName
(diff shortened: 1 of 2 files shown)


danbraunai-goodfire:
We currently have a `spd/app/frontend/src/lib/registry.ts` file which shows the canonical spd and clustering runs. This is probably sufficient for now.

PR #249, opened 2025-11-11
Add flags for better distributed errors

oclivegriffin:
## Description
Adds `NCCL_ASYNC_ERROR_HANDLING` and `TORCH_NCCL_ASYNC_ERROR_HANDLING` flags to give more informative errors in the case of distributed errors.

## Motivation and Context
We've been having weird issues with the new cluster ([example](https://goodfire-ai.slack.com/archives/C08JHTURSQ3/p1762858288831219)). This is just defensive so that we have more information in the case of further failures.

## Related Issue
N/A

## How Has This Been Tested?
manually: `[put run here]`

No

### DIFF
--- a/spd/scripts/run.py
+++ b/spd/scripts/run.py
@@ -161,7 +161,9 @@ def generate_commands(
             mpi_prefix = _build_mpi_prefix(run_id, cmd_idx, dp) if dp > 1 else ""
 
             command = (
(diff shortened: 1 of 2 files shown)


oli-clive-griffin:
@claude 
> The torch version typically wraps the NCCL version
can you check that this is the case in our torch version, and if it is, alter this PR to not duplicate.

PR #252, opened 2025-11-13
fix: Graph freed by pgd loss before backward

oclivegriffin:
## Motivation
Fixes an issue introduced by my recent pr #239. Sloppy from me that I didn't catch this. 

Basically: I refactored the PGD losses to return (loss, n_examples, grad) so that there could be a single central implementation of the gradient calculation to make it consistent. Unfortunately this means that for the final forward pass that we do to report of loss, we also call `torch.autograd.grad` which frees part of the graph. this part of the graph is needed if the loss is used later in `.backward`, and this causes the classic:

```
RuntimeError: Trying to backward through the graph a second time
```

## Description
Rearranges computation so that we don't call `torch.autograd.grad` on the final loss output. Also - disallows multibatch pgd as a loss, as this fix is quite urgent and getting this one working as a loss would be more difficult and also unnecessary for now.

## Related Issue
N/A

## How Has This Been Tested?
Manually. See: https://wandb.ai/goodfire/spd?nw=8f0mb3f1ezy

You can no longer use multi-batch PGD metrics as losses to train on (I don't think anyone has been though)

### DIFF
--- a/spd/metrics/pgd_utils.py
+++ b/spd/metrics/pgd_utils.py
@@ -59,14 +59,14 @@ def pgd_masked_recon_loss_update(
     )
 
     fwd_pass = partial(
-        _pgd_fwd,
-        adv_sources=adv_sources,
+        _forward_with_adv_sources,
+        model=model,
         batch=batch,
+        adv_sources=adv_sources,
         ci=ci,
         weight_deltas=weight_deltas,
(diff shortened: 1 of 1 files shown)


oli-clive-griffin:
@claude added one more small commit, the -1 usage here is correct, right?

PR #260, opened 2025-11-20
Make wandb logs resilient to network outages

oclivegriffin:
## Description
adds `try_wandb` which wraps a wandb (or really any) function and swallows `wandb.errors.CommError`s. As explained in the docstring:

> The choice of wandb CommError is to catch issues communicating with the wandb server but
    not legitimate logging errors, for example not passing a dict to wandb.log, or the wrong
    arguments to wandb.save

Uses this to wrap `wandb.log` and `wandb.save` calls, but not `wandb.init` calls.

## Related Issue
N\A

## Motivation and Context
It's really annoying when a network timeout causes a training run to crash. This allows us to be resilient to temporary network hiccups / slowness at the cost of potential (hopefully small) dead chunks in wandb graphs.

## How Has This Been Tested?
Ran some tms runs locally and turned on and off wifi on my laptop. It worked. Furthermore, I learned that wandb optimises scalar logs really well, saving them to a local cache, so train logs (which are all scalars) (at least in this example of very short local tms runs) actually backfill when internet reconnects. Cool!

No

### DIFF
--- a/spd/utils/wandb_utils.py
+++ b/spd/utils/wandb_utils.py
@@ -454,14 +454,14 @@ def wandb_setup(
 _n_try_wandb_comm_errors = 0
 
 
-def try_wandb[**P, T](fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T | None:
-    """Attempts to call `fn` and if it fails with a wandb CommError, logs a warning and returns
+def try_wandb[**P, T](wandb_fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T | None:
+    """Attempts to call `wandb_fn` and if it fails with a wandb CommError, logs a warning and returns
     None. The choice of wandb CommError is to catch issues communicating with the wandb server but
     not legitimate logging errors, for example not passing a dict to wandb.log, or the wrong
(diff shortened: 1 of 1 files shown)


spd/utils/wandb_utils.py
```diff
@@ -447,3 +449,21 @@ def wandb_setup(
...
+
+def try_wandb[**P, T](fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T | None:
+    """Attempts to call `fn` and if it fails with a wandb CommError, logs a warning and returns
+    None. The choice of wandb CommError is to catch issues communicating with the wandb server but
+    not legitimate logging errors, for example not passing a dict to wandb.log, or the wrong
+    arguments to wandb.save."""
+    global _n_try_wandb_comm_errors
+    try:
+        return fn(*args, **kwargs)
+    except wandb.errors.CommError as e:
```
danbraunai: I'm getting `wandb.errors.errors.AuthenticationError` sometimes, which I think we want to catch too. See [here](https://wandb.ai/goodfire/spd/runs/9hva17te/logs) (you have to download the file to look)

Issue #92, opened 2025-07-28
Make happy path tests use the default configs


danbraunai:
Currently, the happy path tests manually define a config. It would be better if they used the default config for that experiment, and just adjusted  what was needed (C, steps, etc.). This is because we want to know when something breaks with the default config. As it is, there is a risk that something breaking will cause someone to update the config in the tests but not the default configs.

danbraunai-goodfire:
Fixed in #290

PR #264, opened 2025-11-21
Multi-Node Training

oclivegriffin:
## Description
This PR adds multi-node DDP training support and refactors the CLI infrastructure for clarity. We now use torchrun which is the more canonical way to launch distributed jobs, and has the added benefit of not requiring the weird flags we had to add for mpirun to work on the cluster.

Also slightly refactored the distributed utils. We now have a `DistributedState` which is none in the case of single-gpu or cpu based jobs, instead of having sentinel 1-values.

### CLI Split: `spd-simple` and `spd-run`
- **`spd-local`** is a new lightweight CLI for collaborators/public
  - Runs a single experiment at a time via registry name
  - Runs directly without SLURM, git snapshots, or W&B views/reports
  - Supports `--cpu` and `--dp N` flags

- **`spd-run`** has been refactored and streamlined for core team use
  - Removed `--local` flag (always uses SLURM)
  - Always creates git snapshot and W&B workspace view
  - Always uses SLURM

### Cleanup
- Delete `spd/utils/slurm_utils.py` (consolidated into `compute_utils.py`)
- Delete `tests/scripts_run/test_run_spd_cli.py` (tested arg parsing, now we just use `fire`)
- Simplify test suite per testing philosophy: unit test tricky logic, not obvious stuff

## Motivation and Context
We want to train on >8 gpus.

## How Has This Been Tested?
Manually. Waiting for free nodes. Will link runs when I have some going:
See: https://wandb.ai/goodfire/spd/workspace?nw=93r3svowdse

Yes:
- `spd-run --local` removed → use `spd-local` instead
- `spd-run --use_wandb` removed → W&B view is always created
- `spd-run --log_format` removed
- `get_distributed_state()` now returns `None` instead of fake single-process state when not distributed
- Distributed tests now use `torchrun` instead of `mpirun`
- `create_data_loader()` signature changed: `ddp_rank`/`ddp_world_size` → `dist_state: DistributedState | None`

### DIFF
--- a/spd/utils/compute_utils.py
+++ b/spd/utils/compute_utils.py
@@ -109,8 +109,16 @@ def get_command(
             json_tagged_sweep_params = f"json:{json.dumps(sweep_params)}"
             torchrun_cmd += f" --sweep_params_json {shlex.quote(json_tagged_sweep_params)}"
 
-        # Wrap in srun bash -c with proper shell quoting so that $SLURM_PROCID is evaluated on each node
-        command = f"srun bash -c {shlex.quote(torchrun_cmd)}"
+        # Two things here:
+        # 1: We wrap in srun bash -c with proper shell quoting so that $SLURM_PROCID is evaluated on each
+        #    node.
+        # 2: We set --cpus-per-task=128 because:
+        #    Slurm automatically allocates 16 cpus per gpu. We have `DefCpuPerGPU=16` set
+        #    If you run srun bare, it allocates 1 cpu per task. For some reason, we've found that if
+        #    we set --cpus-per-task to anything other than 8 * 16 = 128, we get the error:
+        #    `srun: error: Unable to create step for job 52790: More processors requested than
+        #    permitted`
+        command = f"srun --cpus-per-task=128 bash -c {shlex.quote(torchrun_cmd)}"
         return Command(env_vars=CUDA_FLAGS, command=command)
 
     command += (


spd/app/backend/server.py
oli-clive-griffin: using fire here too so we use fire everywhere

tests/test_distributed.py
```diff
@@ -147,37 +137,32 @@ def _run_experiment(
         self,
         config_path: Path,
         n_processes: int,
-        port: int = 29500,
```
oli-clive-griffin: was never used

spd/scripts/run.py
```diff
@@ -260,378 +247,103 @@ def get_experiments(
...
+        case num_nodes, None:
+            assert num_nodes >= 2, (
+                "if given, num_nodes must be at least 2. pass num_nodes=None to use a single node."
+            )
+            return MultiNode(n_nodes=num_nodes)
+        case _, _:
             raise ValueError(
-                f"DDP (dp > 1) is only supported for lm experiments. "
-                f"Non-lm experiments found: {non_lm_experiments}"
+                "Do not specify both num_nodes and dp. for multi-node runs, assume 8 GPUs per node."
```
danbraunai: How about we just specify `--dp` in the cli and get rid of `--num-nodes`? If dp>8 then you use multiple nodes. If it's too awkward to specify a dp value that isn't divisible by 8, just assert this (though I would be nice to handle if not much effort).

danbraunai: IMO if implementing my suggestion, it'd be nicer to get rid of all the ComputeStrategy stuff and just pass around the dp or n_gpus variable. You'd gate by this variable instead of compute_strategy inside get_command().

spd/scripts/run_simple.py
```diff
@@ -0,0 +1,85 @@
...
+        # Single GPU or CPU
+        cmd = [
+            sys.executable,
+            str(script_path),
+            str(config_path),
+        ]
+
+    if cpu:
+        env_prefix = "CUDA_VISIBLE_DEVICES="
+        logger.info(f"Running: {env_prefix} {' '.join(cmd[:3])} ...")
```
danbraunai:
```suggestion
        logger.info(f"Running: {env_prefix} {' '.join(cmd[:3])}")
```

The dots are confusing

spd/scripts/run_simple.py
```diff
@@ -0,0 +1,85 @@
...
+        subprocess.run(cmd, check=True, env={"CUDA_VISIBLE_DEVICES": ""})
+    else:
+        logger.info(f"Running: {' '.join(cmd[:3])} ...")
```
danbraunai:
```suggestion
        logger.info(f"Running: {' '.join(cmd[:3])}")
```

spd/scripts/run_simple.py
danbraunai: nice

spd/data.py
```diff
@@ -141,13 +142,13 @@ def tokenize_function(
     return tokenized_dataset
 
 
+# TODO docs reflect new args
```
danbraunai: flagging todo

spd/registry.py
```diff
@@ -81,13 +81,6 @@ class ExperimentConfig:
         expected_runtime=60,
         canonical_run=None,
     ),
-    # NOTE: This will be deprecated when we replicate runs with ss_llama_simple
-    "ss_llama": ExperimentConfig(
-        task_name="lm",
-        decomp_script=Path("spd/experiments/lm/lm_decomposition.py"),
-        config_path=Path("spd/experiments/lm/ss_llama_config.yaml"),
-        expected_runtime=2000,
-    ),
```
danbraunai: I think Lucius still does ss_llama runs. I guess I'm OK with this failing hard and him having to manually put it in the registry in the future.

Fyi I added #92 which would handle this (not to be done in this PR though).

danbraunai (commented):
Looked over the rest. All lgtm! Happy for you to merge after:
1. Addressing comments in this and previous review with nothing contenious
2. A claude PR review

I haven't actually tested out running multi-node stuff, there weren't enough nodes available

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dev = [
 
 [project.scripts]
 spd-run = "spd.scripts.run:cli"
+spd-simple = "spd.scripts.run_simple:cli"
```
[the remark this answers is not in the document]
oli-clive-griffin: good idea 👍

spd/data.py
```diff
@@ -141,13 +142,13 @@ def tokenize_function(
...
 def create_data_loader(
     dataset_config: DatasetConfig,
     batch_size: int,
     buffer_size: int,
+    dist_state: DistributedState | None = None,
     global_seed: int = 0,
-    ddp_rank: int = 0,
-    ddp_world_size: int = 1,
```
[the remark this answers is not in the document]
oli-clive-griffin: agreed. should be easier to encapsulate it all under `DistributedState` (or something similar) even in that case though, right?

spd/scripts/run.py
ocg-goodfire: yea good idea. did this

danbraunai-goodfire (approved):
lgtm

PR #276, opened 2025-11-25
[TEST PR] Adding PGD CE Diff metric

leesharkey:
This PR adds a new metric `PGDCEDiff` that computes the cross-entropy (CE) difference between adversarially-optimized PGD-masked model outputs and
   the target model.

   ### Changes:
   - **New file**: `spd/metrics/pgd_ce_diff.py` - Implements the `PGDCEDiff` metric class
   - **Updated**: `spd/configs.py` - Adds `PGDCEDiffConfig` and registers it in `EvalOnlyMetricConfigType`
   - **Updated**: `spd/eval.py` - Adds metric initialization logic and imports
   - **Updated**: `spd/metrics/__init__.py` - Exports the new metric

   ### Key Features:
   - Uses Projected Gradient Descent (PGD) to find adversarial masks that maximize CE loss against true labels
   - Returns metric key `ce_difference_pgd_masked` representing: `CE(pgd_masked_output) - CE(target_output)`
   - Follows the same architecture as existing PGD metrics (`PGDReconLoss`, `PGDReconLayerwiseLoss`)
   - Complements the CE difference metrics in `CEandKLLosses` with PGD-optimized masking
   - Gracefully handles non-LM tasks by detecting output dimensionality (only runs for 3D outputs)

   ### Configuration:
   Add to experiment configs with standard PGD parameters:
   ```yaml
   eval_metric_configs:
     - classname: "PGDCEDiff"
       init: "random"
       step_size: 0.01
       n_steps: 20
       mask_scope: "shared_across_batch"
   ```

   ## Related Issue

   N/A - New feature implementation

   ## Motivation and Context

   During SPD training, we track various metrics to understand how different masking strategies affect model performance. We already have:
   - CE difference metrics for various masking strategies (CI-masked, stochastic, random, etc.) in `CEandKLLosses`
   - PGD reconstruction loss metrics using MSE and KL divergence

   This PR fills the gap by adding a PGD-optimized CE difference metric, allowing us to measure how adversarially-chosen masks (optimized to maximize
   CE loss) compare to the target model. This is useful for:
   - Understanding worst-case masking scenarios
   - Evaluating the robustness of learned component decompositions
   - Measuring maximum possible CE degradation under adversarial component selection

   ## How Has This Been Tested?

   - ✅ Type checking passes (`make type` - basedpyright)
   - ✅ Linting passes (`make format` - ruff)
   - ✅ Pre-commit hooks pass
   - ✅ Code follows existing patterns from `PGDReconLoss` and `CEandKLLosses`
   - ✅ All assertions and error handling in place for fail-fast behavior
   - ✅ **Tested on ResidualMLP experiment**: Metric gracefully skips (returns 0) for non-LM tasks
   - ✅ **Tested on SimpleStories LM experiment**: Metric computes successfully with meaningful values (~20.88 CE difference)

   ### Test Results:
   - ResidualMLP: `eval/ce_kl/ce_difference_pgd_masked: nan` (correctly skipped non-3D output)
   - SimpleStories LM: `eval/ce_kl/ce_difference_pgd_masked: 20.881893157958984` (working correctly!)

   **Note on unit tests**: Following the repository's testing philosophy for research code ("Integration tests often too much overhead for research
   code. Interactive use catches issues at low cost"), no unit tests were added. This is consistent with other PGD metrics in the codebase which also
   lack unit tests. The metric has been validated during actual experiment runs.

   No breaking changes. This is purely additive - a new metric that can be optionally configured in experiment YAML files. Existing experiments
   continue to work without any modifications."


leesharkey:
@claude Please review this PR (specifically the `spd/metrics/pgd_ce_diff.py` and surrounding changes -- Ignore the CLAUDE_ files. Pretend they're not there. Once you're done and have come to a conclusion, cross check your review against the info in the CLAUDE_COMPREHENSIVE.md and CLAUDE_CHECKLIST.md files.

danbraunai (changes requested):
Misc comments:
1. This PR has CLAUDE_CHECKLIST.md and CLAUDE_COMPREHENSIVE.md files in it, we'd want to remove those if merging
2. This PR implements the full-layer version of the loss. We'd want to confirm that we only want this and not also want the subset and layerwise versions of it.
3. Not sure if we actually want this as a Metric that we can use to train on. If we don't want it as a metric, we can probably make it simpler. We may even want to put this the existing ce_and_kl_losses.py file, although I'd make sure that this won't massively slow down that calculation. If it did slow it down, then having it as a separate file that is optional to run makes sense (I suppose maybe we'd want it as a proper "metric" if doing that). Also be mindful if going the ce_and_kl_losses.py route that doesn't reduce over ranks, which is problematic for shared_over_batch.
4. I think we can make minor modifications to the existing pgd functions in pgd_utils.py so that they can return CE or KL. Then we could just call those functions instead of rewriting a lot of the stuff here.

So I think we'll need to chat with Lucius who suggested this feature about some more concrete specs.

PR #279, opened 2025-11-26
Optimize app performance

oclivegriffin:
## Description
- Optimize backend activation context collection with batched numpy processing instead of Python loops
- Switch to columnar data format for more efficient JSON serialization and transfer
- Update frontend components to consume columnar data directly
- Improve progress reporting with normalized 0.0-1.0 floats and throttled updates
- Narrow sidebar to 25vw and enable config text wrapping
- Clean up unused files (package-lock.json, tsconfig splits)

## Related Issue
N/A

## Motivation and Context
The activation contexts app was slow due to inefficient Python loops for collecting top-k examples and verbose nested JSON structures. This PR addresses both backend processing speed and data transfer efficiency.

## How Has This Been Tested?
- Ran the app locally and verified activation context collection works
- Confirmed progress updates stream correctly
- Checked component data loads on demand

No - internal API changes only, no external interfaces affected.

### DIFF
--- a/spd/app/backend/lib/activation_contexts.py
+++ b/spd/app/backend/lib/activation_contexts.py
@@ -112,23 +112,12 @@ class _TopKExamples:
         active_pos = []
         active_ci = []
         for ex in sorted_examples:
-            example_tokens.append(_token_ids_to_strings(ex.window_token_ids, token_strings))
+            example_tokens.append([token_strings[tid] for tid in ex.window_token_ids])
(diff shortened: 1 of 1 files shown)


app/frontend/package-lock.json
oli-clive-griffin: whoops, this was accidentally left behind

.cursor/worktrees.json
oli-clive-griffin: not sure if others want this is here

PR #285, opened 2025-12-03
Attribution local graphs in app

ocg-goodfire:
## Description
Allows for creating and interacting with "local" attribution graphs in the svelte app. I.e. graphs that are built using the ci values of a particular prompt.

## How Has This Been Tested?
Adds end-to-end tests for each api endpoint, except for load_run.

No


danbraunai-goodfire (approved):
Made a few comments but happy to merge provided you write down the things raised in our TODO list.

As for claude's notes:
1. I don't care about CORS allow all for now, since this will be an obvious thing to change if/when we make it public
2. It points out that we use threads which can be risky. My naive guess is that this is fine but could be worth looking into.

Didn't pick up anything else of note (a couple of comments overlap with mine).

PR #288, opened 2025-12-04
Fix loading wandb runs from local filesystem

danbraunai-goodfire:
## Description
- Added parse_wandb_run_path and many unittests to validate and parse wandb paths
- Uses this new parser in RunInfo.from_path
- Refactor RunInfo.from_path to massively reduce the amount of code needed for each subclass

## Motivation and Context
Runs weren't loading from our local filesystem and were always downloading from wandb. Turns out I wrote very lazy code originally which didn't correctly parse the wandb paths.

## How Has This Been Tested?
- Unittests added for parse_wandb_run_path
- No other testing

No


spd/utils/wandb_utils.py
```diff
@@ -29,6 +30,57 @@
...
+        Tuple of (entity, project, run_id)
+
+    Raises:
+        ValueError: If the input doesn't match any expected format.
+    """
+    s = input_path.strip()
+
+    # Strip wandb: prefix if present
+    if s.startswith("wandb:"):
+        s = s[6:]
```
danbraunai-goodfire: Note that this is just optional. Technically, as written, a path of `path/to/X` where `X` is specifically 8 chars and only lowercase letters and numbers, will pass the validation.

If we wanted to enforce `wandb:` prefixes, then we wouldn't be able to parse https:// and we'd have to do other validation when things come in from the app frontend. I'm inclined to just leave this as is, I can't see any security issues apart from killing the app/program.

danbraunai-goodfire: Yeah I'm inclined to remove wandb: altogether. Let me know what you think. A downside is that we'll have things like `pretrained_model_name: goodfire/spd/erq48r3w`, and people will have to infer that it's a wandb path.

PR #289, opened 2025-12-05
Use backend for wandb_path parsing

danbraunai-goodfire:
## Description
Uses the recently added parse_wandb_run_path on the backend to parse the wandb input that the user gives

## Motivation and Context
Currently have complex parsing of wandb paths

No


danbraunai-goodfire:
@claude can you review this PR? Note that your review should cover the scope of this PR only. If you spot things unrelated to this PR, feel free to bring them up and we'll consider them for new issues. I want critical comments, do feel comfortable to give me the hard truths.

Could you check if there are any security issues to think about now that there is no frontend validation before sending to the backend in a GET request?

PR #295, opened 2025-12-10
Show L0 for standard graphs

danbraunai-goodfire:
## Description
- Shows L0 in the view for the standard graph
- Moved the l0_total and l0_per_layer fields to GraphData from OptimizationResult, as its now the same for both standard and optimized graphs.
- Make GraphDataWithOptimization inherit from GraphData to avoid lots of duplication
- Remove l0_per_layer which gets computed and stored but not used anywhere.

image

Yes. Delete your database `rm -rf ~/.spd/local_attr.db*`

### DIFF
--- a/spd/app/backend/schemas.py
+++ b/spd/app/backend/schemas.py
@@ -41,7 +41,7 @@ class GraphData(BaseModel):
     outputProbs: dict[str, OutputProbability]
     nodeImportance: dict[str, float]  # node key -> sum of squared edge values
     maxAbsAttr: float  # max absolute edge value
-    l0_total: float  # total active components at current CI threshold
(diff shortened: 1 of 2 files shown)


spd/app/backend/schemas.py
```diff
@@ -41,6 +41,7 @@ class GraphData(BaseModel):
     outputProbs: dict[str, OutputProbability]
     nodeImportance: dict[str, float]  # node key -> sum of squared edge values
     maxAbsAttr: float  # max absolute edge value
+    l0_total: float  # total active components at current CI threshold
```
ocg-goodfire: int?

danbraunai-goodfire: tyty, fixed

PR #300, opened 2025-12-11
Support models with W_U decompositions

lee-goodfire:
## Description

  Adds support for the decomposed unembedding matrix (`lm_head`) in the SPD app visualization. The unembedding matrix components now appear in the attribution graph between the final down_proj layer and the output tokens, displayed as "W_U" in the UI.

  Changes:
  - **Backend**: Conditionally include `lm_head` in the layer list when the model has a decomposed unembedding matrix
  - **Frontend**: Add `lm_head` to layer ordering, parsing, and display logic; show as "W_U" in all user-facing labels

  ## Related Issue

  N/A

  ## Motivation and Context

  Previously, the app only supported models without decomposed unembedding matrices. With the addition of unembedding matrix decomposition to the SPD framework, we needed the app to visualize these components alongside the other decomposed layers.

  ## How Has This Been Tested?

  - Loaded a model with decomposed unembedding matrix (WandB run `zg2zjtq1`) and verified:
    - W_U layer appears in the correct position (after down_proj, before output)
    - Edges connect correctly to/from W_U components
    - W_U components can be selected and intervened on
    - Tooltips and pinned nodes display "W_U" correctly
  - Verified backward compatibility with models that don't have `lm_head` decomposition
  - Frontend type checking and linting pass

  No. Models without decomposed unembedding matrices continue to work as before.

### DIFF
--- a/spd/app/backend/compute.py
+++ b/spd/app/backend/compute.py
@@ -15,7 +15,11 @@ from torch import Tensor, nn
 from tqdm.auto import tqdm
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
 
-from spd.app.backend.optim_cis.run_optim_cis import OptimCIConfig, optimize_ci_values
+from spd.app.backend.optim_cis.run_optim_cis import (
+    OptimCIConfig,
+    compute_label_prob,
+    optimize_ci_values,
+)
 from spd.configs import SamplingType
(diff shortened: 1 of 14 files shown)


ocg-goodfire (approved):
Oh nice. thought this would end up being gross but looks clean

PR #302, opened 2025-12-11
Sort node by ci instead of edge importance

danbraunai-goodfire:
## Description
- Remove nodeImportance everywhere
- Remove the Node Layout option (minor note that jitter used to mix up the q/k/v).
- Sort nodes by their CI value
- Show the CI value in the box when hovering over a node
- Add node opacity based on CI values
- Backend refactor to separate out the streaming logic from the graph computation and post-processing logic

NOTE: I've still kept around the GLOBAL_EDGE_LIMIT. We no longer have the issue that if edges disappear, nodes disappear too, because nodes will appear if they are above the ci_threshold regardless. This will fix the need for Lucius increasing the edge limit to ensure that more nodes appeared. I think the GLOBAL_EDGE_LIMIT is now a decoupled issue.

image

## How Has This Been Tested?
Manual look in the app

No

### DIFF
--- a/spd/app/backend/routers/graphs.py
+++ b/spd/app/backend/routers/graphs.py
@@ -59,7 +59,9 @@ class GraphData(BaseModel):
     tokens: list[str]
     edges: list[EdgeData]
     outputProbs: dict[str, OutputProbability]
-    nodeCiVals: dict[str, float]  # node key -> CI value (causal importance)
+    nodeCiVals: dict[
+        str, float
+    ]  # node key -> CI value (or output prob for output nodes or 1 for wte node)
     maxAbsAttr: float  # max absolute edge value
     l0_total: int  # total active components at current CI threshold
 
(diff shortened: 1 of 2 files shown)


spd/app/backend/routers/graphs.py
```diff
@@ -459,158 +460,117 @@ def compute_graph_optimized_stream(
...
-    return StreamingResponse(generate(), media_type="text/event-stream")
+    wte nodes get CI=1.0 (always visible), output nodes use their probability.
+    """
+    result = dict(node_ci_vals)
+    for seq_pos in range(num_tokens):
+        result[f"wte:{seq_pos}:0"] = 1.0
+    for key, output_prob in output_probs.items():
+        seq_pos, token_id = key.split(":")
+        result[f"output:{seq_pos}:{token_id}"] = output_prob.prob
+    return result
```
danbraunai-goodfire: So yes, I've committed the sin here of "pretending" that wte and output layers have ci values. The wte isn't much of a stretch, but the output is stretching a little bit. I've done this so that it's simpler filtering edges (we just include edges that connect to a node_ci_vals_with_pseudo) and for sorting.

We could change the name to "node_value" or something like that, which is a ci_value for all but the output nodes, but since they're nearly all ci values I didn't do this, and just commented on the frontend schema.

danbraunai-goodfire: I'm inclined to leave it for now. Bit of effort and mess to add types for input, ci_val, and output separately on both the front and backend. Certainly not against someone doing this themselves though, I just don't weigh the benefits big enough.

PR #303, opened 2025-12-11
Target Model Probabilities in Intervention Run History

lee-goodfire:
## Description
Add target model token probabilities to the intervention run history in the SPD app. When viewing intervention results, users now see both SPD model predictions and target model baseline probabilities side-by-side, displayed as "SPD: X.X%" and "Targ: X.X%" for each predicted token.

## Related Issue
  N/A

## Motivation and Context
When analyzing intervention results, it's useful to compare how the SPD model's predictions differ from the target model's baseline. Previously, only SPD probabilities were shown, making it difficult to gauge whether a component's activation significantly changes the model's behavior. This change enables direct comparison.

## How Has This Been Tested?
  - Backend type checking with basedpyright (0 errors)
  - Frontend type checking with svelte-check (0 errors)
  - Linting with ruff (all checks passed)
  - Unit tests for the intervention forward pass logic
  - Edge case tests for boundary conditions
  - I've (Lee) also checked that it works in the app (User test)

Yes - the `TokenPrediction` schema now requires a `target_prob` field. Existing intervention runs stored in the database will fail to deserialize since they lack this field. 

Options to address:
  1. Clear existing intervention runs before upgrading
  2. Add a migration to backfill `target_prob: 0.0` for existing records
  3. Make `target_prob` optional with a default (not recommended - hides missing data)

### DIFF
--- a/spd/app/backend/compute.py
+++ b/spd/app/backend/compute.py
@@ -649,11 +649,11 @@ def compute_intervention_forward(
     with torch.no_grad():
         # SPD model forward pass (with component masks)
         spd_logits: Float[Tensor, "1 seq vocab"] = model(tokens, mask_infos=mask_infos)
-        spd_probs = torch.softmax(spd_logits, dim=-1)
+        spd_probs: Float[Tensor, "1 seq vocab"] = torch.softmax(spd_logits, dim=-1)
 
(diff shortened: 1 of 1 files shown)


ocg-goodfire (approved):
LGTM!

PR #363, opened 2026-02-06
Makes ComponentModel generic over batch and output types

danbraunai-goodfire:
## Description

TODO (either here or afterwards):
- Currently in spd we use n_ctx=512 for the dataset config and not n_ctx=513 which we use for pretrain. The spd code will calculate ce losses using the fancy ignore_index=-100 thing. Once we sort out this PR, I think we should make our dataloaders return batch and labels, and then we can set n_ctx for the dataset to 513 just like we do for pretrain, and not do this fancy ignore_index stuff. We should also rename max_seq_len in the spd task_config to n_ctx.


ocg-goodfire:
@danbraunai-goodfire I think I definitely went overboard with the abstractions so thanks for reigning this is lol. I do have some suggestions in https://github.com/goodfire-ai/spd/pull/375. lmk what you think

danbraunai-goodfire:
Closing because this should be pointing to dev

PR #407, opened 2026-02-19
Add topological interpretation module

ocg-goodfire:
## Description

New `spd/topological_interp/` module for context-aware component labeling using network graph structure. Unlike standard autointerp (one-shot per component), this uses dataset attributions to provide neighbor context in each component's prompt.

**Three-phase pipeline:**
1. **Output pass** (late→early): "What does this component DO?" — each prompt includes top-K downstream neighbors by attribution, with their labels
2. **Input pass** (early→late): "What TRIGGERS this component?" — each prompt includes top-K upstream neighbors + co-firing components (Jaccard/PMI)
3. **Unification**: synthesizes output + input labels into a single unified label

Output and input passes are fully independent (both layer-serial within, but no cross-dependency). Unification combines them.

**Also includes:**
- Extraction of shared prompt helpers from `dual_view.py` into `spd/autointerp/prompt_helpers.py` (public API, used by both autointerp and topological interp)
- Layer ordering via topology module's `CanonicalWeight` system (correct for all model architectures)
- `spd-topological-interp` CLI entry point
- SQLite DB with WAL mode storing output/input/unified labels + directed prompt edge graph
- Resume support per-phase via completed key sets

## Motivation and Context

Current autointerp labels each component independently. This misses the graph structure — components influence each other through the network, and knowing what upstream/downstream components do provides crucial context for interpretation. Inspired by Curt Tigges' topological labeling approach (`.circuits-ref/`).

## How Has This Been Tested?

- basedpyright: 0 errors across all new + modified files
- ruff: lint + format clean
- All 442 existing tests pass (0 regressions from `dual_view.py` refactor)
- Ordering logic verified with unit-style assertions across attn/mlp/glu layer types
- `spd-topological-interp --help` works
- All module imports verified end-to-end

No. `dual_view.py` functions are re-exported from the same module (now importing from `prompt_helpers.py`). All existing autointerp functionality is preserved.


ocg-goodfire:
Already in dev from #427

PR #410, opened 2026-02-22
Speed up dataset attribution harvesting ~31x

ocg-goodfire:
## Description

Dataset attribution harvesting was prohibitively slow on larger decompositions (s-17805b61: 39K components, 50K vocab). This PR makes it ~31x faster and adds convergence analysis showing we need far fewer batches than the current default.

### What changed

**1. Vectorize accumulation (14x speedup)**

The inner loop `for c in alive: target_col[start + c] += attr[c]` ran ~7K Python iterations per backward pass. Replaced with a single `scatter_add_` call — the same pattern already used for wte sources three lines above. This was the actual bottleneck, not the backward passes themselves (discovered by profiling).

**2. Input-residual trick (2.3x speedup)**

For target layers where `d_input < C`, we backprop from `d_input` dimensions instead of `C`, then recover per-component attributions via `accumulator @ V` at query time. Reduces backward passes from 39K to 26K on s-17805b61. Analogous to the existing output-residual trick for the unembed layer.

**3. Refactor harvester internals**

- Unified `_process_component_targets_direct` and `_process_component_targets_residual` into a single `_process_targets` method
- Made `comp_accumulator` private, exposed via `get_comp_attributions()` (no more incomplete public state)
- Persistent hooks with `_capturing` flag instead of register/remove per batch
- Pre-computed `_TargetLayerInfo` dataclass per layer

**4. Split CLI into run_worker.py and run_merge.py**

Mirrors the `spd.harvest` script structure. The old `run.py` combined both behind a `--merge` flag, requiring a dummy `config_json` positional arg in merge mode.

**5. Convergence analysis tooling** (new scripts, not in production path)

- `convergence_merge_gpu.py`: GPU-accelerated incremental merge that computes per-target top-k source overlap vs final at each prefix. Two-pass to avoid OOM.
- Results visualized at `SPD_OUT_DIR/www/attribution_convergence.html`

### Convergence findings

On s-eab2ace8 (7K components, 160-point curve from 13 to 2000 batches):
- **Mean top-4 overlap with final >99%** by ~100 batches
- **500 batches** gives >99.7% mean overlap — the default of 10,000 is ~20x overkill
- Even the worst single target has >99% overlap by ~500 batches

s-17805b61 and s-6a9dcb7c convergence runs in progress.

## How Has This Been Tested?

- Smoke test: 1-batch harvest on s-eab2ace8, correct output shape and nonzero count
- A/B benchmark: old (67s/batch) vs new (2.2s/batch) on s-eab2ace8, batch_size=32
- Checksum comparison: 1.2% relative error (bf16 precision in V projection)
- Production runs: 40 GPU x 25 batches for s-17805b61 and s-6a9dcb7c (in progress)

- `spd.dataset_attributions.scripts.run` split into `run_worker` and `run_merge`
- `harvester.finalize_comp_accumulator()` renamed to `harvester.get_comp_attributions()`
- The SLURM launcher (`run_slurm.py`) is updated to use the new scripts


ocg-goodfire:
@claude pls confirm these changes (in spirit) are already on the branch `dev`

ocg-goodfire:
For posterity, the above is wrong

in `spd/dataset_attributions/harvester.py`:

Dev (line 273-274):
```python
ci_weighted = (grad * act * ci[source_layer]).sum(dim=(0, 1))  # full vector
regular_acc[target_layer][source_layer][t_idx].add_(ci_weighted)
```

here (line 263-264):
```python
attr = (grad * act).sum(dim=(0, 1))
target_col.scatter_add_(0, start + alive, attr[alive])
````
That's the biggest optimization here we've already got in dev. the input-residual trick is incompatible with attributions to absolute values, so was dropped

PR #419, opened 2026-03-03
Transcoder harvest integration

ocg-goodfire:
# Description
Ports @bartbussmann 's PR https://github.com/bartbussmann/spd/pull/7 to this repo with a few small changes to make it seamlessly integrate with the harvest CLI


ocg-goodfire:
@claude can you please review

ocg-goodfire:
Addressed the valid points from the review:

- **Dataloader duplication**: Extracted `pretrain_dataloader()` helper in `spd/adapters/base.py` — both CLT and Transcoder adapters now call it (077766ee)
- **Type hint**: Added explicit `list[torch.utils.hooks.RemovableHandle]` to TranscoderHarvestFn hooks
- **Merged dev** into this branch (picks up the multi-provider LLM support from #446)

Re: the other suggestions — artifact download robustness (checking config.json) isn't worth the complexity; if the download is partial it'll fail on load anyway. The BatchTopK assertion is intentional — it's the only encoder type used in practice.

Ready for re-review / merge.

PR #440, opened 2026-03-13
Fix KL divergence eval OOM: use fused reduction

ocg-goodfire:
## Summary

`calc_kl_divergence_lm` used `F.kl_div(reduction="none")` which materializes a full `[batch, seq, vocab]` intermediate tensor (~13GB with eval_batch_size=128, seq=512, vocab=50K). Called 6x per eval step in `CEandKLLosses`, this was the memory high-water-mark causing OOMs on runs with many components (e.g. the C_4x feature-splitting run with 38K components).

Fix: use `reduction="sum"` which fuses the reduction into the CUDA kernel, then divide by `n_positions` to match the original mean-over-positions semantics. Verified numerically equivalent via `torch.testing.assert_close`.

## Test plan

- [x] All local tests pass
- [x] basedpyright clean on changed file
- [x] Numerical equivalence verified: `F.kl_div(reduction="sum") / n_positions` matches `F.kl_div(reduction="none").sum(-1).mean()`
- [ ] CI passes

### DIFF
--- a/spd/utils/general_utils.py
+++ b/spd/utils/general_utils.py
@@ -209,27 +209,16 @@ def extract_batch_data(
 def calc_kl_divergence_lm(
     pred: Float[Tensor, "... vocab"],
     target: Float[Tensor, "... vocab"],
-    reduce: bool = True,
-) -> Float[Tensor, ""] | Float[Tensor, "..."]:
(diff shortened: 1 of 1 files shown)


ocg-goodfire:
Superseded by new PR with cleaner branch name.

PR #441, opened 2026-03-13
Add data presentation context to autointerp prompt

ocg-goodfire:
## Description

Adds a "Data Presentation" section to the autointerp prompt template that explains how activation examples are constructed — specifically that they are fixed-size token windows extracted from longer sequences, with boundary behavior at sequence edges.

## Motivation and Context

The autointerp LLM had no context about how the data it was analyzing was generated. For example, when a component consistently fired at early token positions, the interpreter couldn't recognize this because it didn't know that a short left context in an example implies the firing was near the start of a sequence. This led to missed positional patterns in labels.

## How Has This Been Tested?

Type-checked with basedpyright (0 errors). Ruff clean.

No — `ArchitectureInfo` gains two new required fields (`seq_len`, `context_tokens_per_side`) but all construction sites are updated. Existing harvest data is compatible (reads `context_tokens_per_side` from the already-saved `config.json`).

### DIFF
--- a/spd/app/backend/routers/correlations.py
+++ b/spd/app/backend/routers/correlations.py
@@ -170,8 +170,11 @@ async def request_component_interpretation(
             detail="OPENROUTER_API_KEY environment variable not set",
         )
 
+    from spd.harvest.schemas import get_activation_contexts_dir
+
     # Get architecture info and tokenizer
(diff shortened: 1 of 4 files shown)


ocg-goodfire:
Superseded — rebased onto dev as #442 (main was stale).

PR #456, opened 2026-03-18
Refactor clustering to use compressed memberships

ocg-goodfire:
[entirely by codex 5.4]

## Summary
- replace the dense sample-by-group thresholded clustering path with exact compressed memberships
- keep memberships hybrid sparse/bitset so low-density groups stay sparse and denser merged groups upgrade naturally
- route thresholded clustering runs through the compressed path, add multi-batch non-LM collection via `n_samples`, and keep the old dense path as a fallback when `activation_threshold` is `None`
- speed up exact merge compute by moving dense MDL cost evaluation to GPU and replacing Python membership-intersection overlap updates with a row-oriented `numba` kernel over a cached sample-by-component CSR matrix
- add a snapshot/benchmark scaffold so activations can be harvested once and merge kernels can be benchmarked repeatedly without recollecting data

## What Changed
The original exact thresholded clustering path materialized dense sample-by-group state and then recomputed merged-group overlaps with Python-level set operations. That made memory scale badly with token count, and once memory was fixed, the per-merge overlap update became the main runtime bottleneck.

This PR changes the exact thresholded path in two stages:

1. Representation changes
- introduce exact compressed memberships for thresholded clustering
- store each group as sparse sample indices when that is cheaper than a bitset, otherwise as a packed bitset
- build exact coactivation matrices from the compressed memberships with sparse `X.T @ X`
- add multi-batch collection for non-LM clustering with `n_samples`

2. Exact runtime changes
- keep the dense coactivation matrix on GPU when possible so `compute_merge_costs()` runs there
- build the original sample-by-component CSR matrix once at merge start
- after each merge, recompute merged-group overlaps by scanning only the merged rows through a `numba`-compiled row-oriented kernel instead of doing Python-level membership intersections against every current group

Semantics are still exact for the thresholded path. The new row-oriented overlap update computes the same counts as the old direct membership intersection path; it just does it with compiled array code over the original component activity matrix and current group assignments.

## Performance Notes
Jose-like synthetic benchmark matching the old real run scale after dead filtering:
- `500,000` sampled tokens
- `9,981` alive components
- token-wise `l0 ~= 200`

Measured progression:
- original compressed merge loop: about `1.90s/iter`
- after GPU cost-matrix move: about `0.66s/iter`
- after row-oriented `numba` overlap update: about `0.095s/iter`

Current Jose-like exact runtime at that scale:
- CSR build: about `2.5s`
- coactivation build: about `75s`
- merge loop: about `0.095s/iter`
- projected `~9100` iterations: about `16 minutes` merge-side total at `500k` sampled tokens

That means the exact merge kernel is now roughly:
- `~20x` faster than the original compressed path
- `~6.9x` faster than the previous branch state before the row-oriented update

On the memory side, the compressed membership representation already reduced token-scaling memory dramatically versus the old dense thresholded path, so the current branch materially improves both:
- memory headroom for larger token counts
- merge throughput at Jose-like clustering scale

## Validation
- `python3 -m compileall spd/clustering`
- `uv run pytest tests/clustering/test_merge_integration.py tests/clustering/test_filter_dead_components.py tests/clustering/test_merge_pair_samplers.py`
- `uv run pytest tests/clustering/test_merge_integration.py -q`
- targeted dense-vs-compressed merge sanity check with synchronized sampler state
- targeted sanity check that low-density memberships stay sparse
- targeted exactness check that the row-oriented overlap recompute matches the previous direct membership recompute
- Jose-like synthetic benchmarks for coactivation build, overlap-update subproblem, and full merge-loop throughput


ocg-goodfire:
@claude Please review this. mostly laser-focus on correctness - this should be an interface-preserving efficiency improvement.

ocg-goodfire:
@claude thanks for you review. Can you please review the 2 new commits since? again - mostly concerned with correctness.

ocg-goodfire:
@claude sorry one more how's that last commit look

ocg-goodfire:
@claude can you please summarise this PR? I promise I didn't just get codex to do the whole thing and have little understanding about exactly how it works... 👀

ocg-goodfire:
split into 3 PRs:
- https://github.com/goodfire-ai/spd/pull/463
- https://github.com/goodfire-ai/spd/pull/464
- https://github.com/goodfire-ai/spd/pull/465

so closing this one

PR #463, opened 2026-03-28
Clustering core: compressed memberships, harvest-merge split

ocg-goodfire:
## Summary

Overhaul of the clustering module: compressed membership representation, harvest-merge split, and architectural cleanup.

### Compressed memberships
- **`CompressedMembership`** (`sample_membership.py`): sparse/bitset storage for per-component activation patterns, with numba JIT for set intersection and union operations
- **`MembershipBuilder`** (`memberships.py`): streaming builder that accumulates thresholded boolean memberships from GPU batches without materializing the full dense `[n_samples, n_components]` matrix
- **`ProcessedMemberships`** (`memberships.py`): the core data type with save/load (sparse CSC + metadata JSON + optional preview tensor), replacing the old dense `ActivationsTensor` path and the intermediate `MembershipSnapshot` type
- Removed the uncompressed (dense tensor) merge codepath entirely — `merge_iteration` (dense) deleted, only `merge_iteration_memberships` (compressed) remains

### Harvest-merge split
- **`spd-cluster-harvest`** (`run_harvest.py`): GPU activation collection, saves `ProcessedMemberships` to disk
- **`spd-cluster-merge`** (`run_merge.py`): CPU-only merge from saved snapshot, with optional log callback for wandb
- **`run_clustering.py`**: thin orchestrator that calls `harvest()` then `merge()` with wandb logging — single implementation path, no duplicated logic
- Ensemble pipeline (`spd-clustering`) pre-generates run IDs and passes them via CLI args, following the same pattern as the harvest module

### Config consolidation
- **`HarvestConfig`** owns all data collection + filtering params (activation_threshold, filter_dead_threshold, filter_dead_stat, module_name_filter)
- **`MergeConfig`** is purely merge algorithm params (alpha, iters, sampling method)
- **`ClusteringRunConfig`** embeds `harvest: HarvestConfig` + `merge: MergeConfig` instead of duplicating fields
- `collect_memberships()` takes `HarvestConfig` directly (was 12 separate args)

### Architectural cleanup
- **Deleted `ensemble_registry.py`**: replaced SQLite DB with pre-assigned run IDs passed via CLI args
- **Deleted `storage.py`**: replaced `StorageBase`/`ExecutionStamp` indirection with direct path computation
- **Added `paths.py`**: centralized output directory functions + ID generation (`new_run_id()`, `new_harvest_id()`, `new_ensemble_id()`)
- **Added `get_command()` pattern**: `run_clustering.py` and `calc_distances.py` expose command builders, pipeline calls those instead of inline string assembly
- **Extracted `memberships.py`** from `activations.py`: clean separation of compressed membership machinery from dense activation utilities

### Other
- `exp_rank` merge pair sampler
- `numba` dependency (replaces `hf-transfer`)
- `activation_threshold` on `MergeConfig` removed (now lives on `HarvestConfig` where it belongs)

## Test plan
- [x] All clustering tests pass (57 passed, 1 skipped)
- [x] basedpyright clean, ruff clean
- [x] E2E smoke test on Jose (s-55ea3f9b): all 3 entry points (harvest, merge, combined) produce identical results

Generated with [Claude Code](https://claude.com/claude-code)


ocg-goodfire:
@claude please review this, the internals have changed quite a bit but the e2e behaviour should be mostly the same. Please taxonomize the large internal mechanical differences, and pay close attention to any correctness issues in core code.

ocg-goodfire:
Thanks for the thorough review! This is Claude (Opus, the one who wrote the code) responding — Oli asked me to reply.

Regarding the three warnings:

1. **`np.bitwise_or.at()` race condition** — agreed this is a non-issue. Duplicate indices would OR the same bit twice, which is idempotent. No data corruption possible.

2. **Memory pressure on coactivation matrices** — handled by `_choose_coact_device()` in `merge.py` which checks free GPU memory before transferring. Falls back to CPU if insufficient. For Jose (5k components) the coactivation matrix is ~100MB, well within bounds.

3. **Float32 precision in costs** — the merge algorithm only cares about ordering (which pair has lowest cost), not absolute values. Float32 is plenty for that.

Merging now!

PR #467, opened 2026-04-03
Remove hardcoded goodfire entity from SPDAdapter

Antovigo:
## Description
`SPDAdapter` now accepts a full `wandb_path` instead of just a run ID, removing the hardcoded `goodfire/spd/runs/` prefix.

## Motivation and Context
The harvest pipeline was unusable for anyone with a different wandb entity because `SPDAdapter.__init__` only took a run ID and hardcoded `goodfire/spd/runs/{run_id}`. Now the full wandb path flows through from the harvest config.

## How Has This Been Tested?
- `make check` passes (basedpyright + ruff)
- Tested harvest submission on a non-goodfire wandb entity

No. `adapter_from_config` still works the same way (reads `wandb_path` from `SPDHarvestConfig`). `adapter_from_id` now always goes through `_load_method_config`, which reads the full config from the harvest DB. This requires harvest data to exist, but `adapter_from_id` is only called from downstream steps (autointerp, intruder eval, graph_interp) that already depend on harvest.

### DIFF
--- a/spd/adapters/__init__.py
+++ b/spd/adapters/__init__.py
@@ -23,7 +23,7 @@ def adapter_from_config(method_config: DecompositionMethodHarvestConfig) -> Deco
         case SPDHarvestConfig():
             from spd.adapters.spd import SPDAdapter
 
-            return SPDAdapter(method_config.id)
+            return SPDAdapter(method_config.wandb_path)
         case TranscoderHarvestConfig():
             from spd.adapters.transcoder import TranscoderAdapter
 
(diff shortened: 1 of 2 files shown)


spd/adapters/__init__.py
```diff
@@ -41,11 +41,6 @@ def adapter_from_id(decomposition_id: str) -> DecompositionAdapter:
     method config from the harvest DB (which is always populated before downstream
     steps like autointerp run).
     """
-    if decomposition_id.startswith("s-"):
```
Antovigo, line 44: I'm not 100% sure what these lines were for. Are there special cases when one might want to use a SPDAdapter without running the harvest first?

PR #474, opened 2026-04-21
Add RunBatch and ReconstructionLoss Protocols

danbraunai-goodfire:
## Description
Introduces `RunBatch` and `ReconstructionLoss` protocols (`spd/models/batch_and_loss_fns.py`) that parameterize `ComponentModel` and training for different target model architectures.

- `ComponentModel.__init__` now takes a `run_batch: RunBatch` callable instead of a `pretrained_model_output_attr` string. For simple cases, `make_run_batch(output_extract)` builds one from `None` / int / str (passthrough / tuple index / attribute name).
- `optimize()` takes a `reconstruction_loss: ReconstructionLoss` callable (`recon_loss_mse`, `recon_loss_kl`) instead of reading `output_loss_type` from config.
- Removes `extract_batch_data`; batch movement and output extraction now flow through `RunBatch`.
- Config migration: `pretrained_model_output_attr="idx_0"` → `output_extract=0`, `pretrained_model_output_attr="logits"` → `output_extract="logits"`. `output_loss_type` is dropped.

## Motivation and Context
`ComponentModel` previously assumed the batch was a single tensor and the target output was a tensor reachable via a single attribute/index string. This blocked target models that take structured inputs (e.g. IH passes `(tokens, labels)`). Replacing the config knobs with injected callables removes those assumptions and keeps the config schema small.

## How Has This Been Tested?
- `make check` and `make test` pass
- Updated tests in `tests/test_component_model.py`, `tests/test_spd_losses.py`, metrics tests
- New `tests/test_batch_and_loss_fns.py`
- Ran `spd-local` sanity check on `tms_5-2` and IH configs

Yes. Old checkpoints with `pretrained_model_output_attr` / `output_loss_type` fields are auto-migrated by the config loader, but any external code instantiating `ComponentModel` directly must now pass `run_batch` instead of `pretrained_model_output_attr`, and any code calling `optimize()` must pass `reconstruction_loss`.

## Follow-ups
Several items from the PR review are deliberately not addressed here and are tracked in #475 — three app-side correctness bugs (canonical-key refactor consumers, hardcoded dataset, `get_tokenized_results` key mismatch) plus the declined/out-of-scope @claude comments. They'll be picked up in a follow-up PR.


danbraunai-goodfire:
@claude can you review this PR? Note that your review should cover the scope of this PR only. If you spot things unrelated to this PR, feel free to bring them up and we'll consider them for new issues. I want you to think about possible ways that the implementation could be improved. I want critical comments, do feel comfortable to give me the hard truths.

danbraunai-goodfire:
Comments from a claude "ultrareview":
"""
🟡 spd/app/backend/routers/autointerp_compare.py:61-68
The timestamp parser in _parse_subrun_timestamp uses the format %Y%m%d_%H%M%S but subrun IDs are actually created with a- + strftime("%Y%m%d_%H%M%S_%f") (see spd/autointerp/scripts/run_interpret.py:55 and run_slurm.py:34), so strptime raises ValueError: unconverted data remains: _123456 on every real subrun ID. The except branch swallows the error and returns the raw string, so the UI shows 20260422_120000_123456 instead of the intended 2026-04-22 12:00:00. Fix by using format %Y%m%d_%H%M%S_%f.

🟡 CLAUDE.md:341-346
CLAUDE.md documents spd-postprocess as taking a wandb path (spd-postprocess <wandb_path>) and a --config flag, but the CLI actually takes a single positional argument that must be a path to a PostprocessConfig YAML file — and there is no --config flag. A user (or AI agent) following the docs with spd-postprocess wandb:ent/proj/run will hit a FileNotFoundError in PostprocessConfig.from_file(). Fix by updating the docs to spd-postprocess <config.yaml> (matching the module's own docstring at spd/postprocess/cli.py:8-10), or extending the CLI to accept a wandb path and synthesize a default config.

🔴 spd/app/backend/compute.py:290-319
The canonical-key refactor renamed frontend-returned node keys from concrete paths (e.g. h.0.mlp.c_fc:5) to canonical (0.mlp.up:5), but two consumers of those keys still expect the old concrete form and were never updated. (1) filter_ci_to_included_nodes in spd/app/backend/compute.py asserts layer names exist in ci_lower_leaky.keys(), which are concrete paths — so the assertion fires for every non-empty selection, fully breaking the "generate subgraph from selection" feature in the Prompt Attributions tab. (2) _tool_run_ablation in spd/app/backend/routers/mcp.py (lines 852–860) appends raw canonical layers to active_nodes; compute_intervention then KeyErrors on ci_masks[layer] because that dict is keyed by concrete paths from model.module_to_c. Fix: call loaded.topology.canon_to_target(layer) before validating/using the layer, mirroring intervention.py:_parse_node_key and the other MCP tools (probe_component, get_component_info, get_component_activation_examples) which already do this correctly.

🔴 spd/app/backend/routers/dataset_search.py:414-432
Two new endpoints (get_random_samples_with_loss in dataset_search.py:414-432 and _tool_search_dataset in mcp.py:896-936) hardcode lennart-finke/SimpleStories and the story column, ignoring the loaded run's LMTaskConfig.dataset_name/column_name. The PR explicitly adds pile_llama_simple_mlp-{2L,4L,12L} registry entries whose dataset is danbraunai/pile-uncopyrighted-tok-shuffled — the dataset-random UI panel (DatasetRandomPanel.svelte:16 fires getRandomSamplesWithLoss on mount) silently shows SimpleStories samples scored by the pile_llama model for those runs. Fix: read dataset_name and column_name from loaded.config.task_config like the main /api/dataset/search endpoint does.

🟡 spd/app/backend/routers/graphs.py:734
spd/app/backend/routers/graphs.py:734 uses Query(gte=0) for imp_min_coeff, but gte is not a recognized Pydantic Query kwarg (the valid ones are ge/le/gt/lt). Unknown kwargs are silently absorbed into json_schema_extra rather than installing a validator, so negative imp_min_coeff values pass validation and reach the optimizer — where a negative coefficient flips the importance-minimality term into a reward for more active components. Fix: change gte=0 to ge=0 (every other numeric query param in the same function already uses the correct ge=/gt= spelling).

🔴 spd/app/backend/routers/dataset_search.py:249-268
The new get_tokenized_results endpoint (line 262) reads result["story"], but search_dataset (lines 141-147) stores results with the key "text" — so every call to GET /api/dataset/results_tokenized after a search raises KeyError: 'story'. The metadata filter at line 287 has the same mismatch (excludes ["story", "occurrence_count"] but the keys are ["text", "occurrence_count", "metadata"]), so the story text leaks into the output metadata and the inner metadata dict gets stringified instead of spread. Fix by using result["text"] and either spreading result["metadata"] or filtering on ["text", "occurrence_count", "metadata"].
"""

PR #479, opened 2026-04-26
Intruder eval rework + per-provider rate limits + comparison export

ocg-goodfire:
## Description

Three commits, each independently reviewable:

### 1. `6c6b9ff4` Intruder eval rework + autointerp provider rate limits

**Intruder (`spd/harvest/intruder.py`):**
- **Streaming trial generation.** `_build_trials` is now a lazy iterator that fetches one component at a time from the harvest DB. Old behaviour loaded all `ComponentData` upfront (heavy, doesn't scale).
- **Lightweight `DensityIndex`.** Stores only `(component_key, firing_density)` tuples instead of full `ComponentData`. New `db.get_component_densities(min_examples)` does this in one SQL query — no JSON blob deserialization.
  - Has a fallback for harvest DBs without the `n_activation_examples` column (uses `json_array_length(activation_examples)`). Mildly inconsistent with the project's no-legacy stance but small and defensible for shared NFS DBs.
- **Behaviour change:** if no donor lies within `density_tolerance`, the trial is now **skipped** (was: widened to all components). Affects which components get scored at the tails of the density distribution.
- **XML prompt format** (`<raw>` + `<annotated>` views) replaces the old `<<delimiter>>` format. JSON response with `reasoning` field, persisted on `IntruderTrial`.
- **Prompts saved** to a new `intruder_prompts` table (`HarvestDB.save_intruder_prompt`).
- API: `run_intruder_scoring` now takes `db: HarvestDB` instead of separate `components: list[ComponentData]` + `score_db`.
- `max_tokens` 300 → 4000 (reasoning field).
- New SLURM CLI: `spd-intruder <decomposition_id> <harvest_subrun_id>`.
- `IntruderSlurmConfig.time` 4h → 10h.

**Per-provider rate limits:**
- `max_concurrent` / `max_requests_per_minute` moved from `AutointerpConfig`, `AutointerpEvalConfig`, `GraphInterpConfig`, `IntruderEvalConfig` onto each `*LLMConfig` variant in `spd/autointerp/providers.py`. Defaults reflect each provider's actual limits (Opus 20/100, Sonnet 40/300, Google 100/1000, etc.).
- All four call sites updated: `config.max_concurrent` → `config.llm.max_concurrent`.

### 2. `4f377fce` Remove `_ensure_table` migration shim

Pure deletion of dead migration code in `HarvestDB`.

### 3. `574d400f` Add intruder score comparison data export

`spd/harvest/scripts/compare_intruder_scores.py` reads runs listed in `scripts/intruder_comparison.json` and writes:
- `data.json` — per-model `{decomposition_id, harvest_subrun_id, densities[], scores[], summary}` + group means + palette spec
- `README.md` — schema, summary tables, intended-plot description

Output: `SPD_OUT_DIR/intruder_comparison/`. Decoupled from plotting so visuals can be iterated downstream without rerunning.

## Motivation and Context

Original branch was `intruder-rework`, last touched late March. Rebased onto current `dev` (with the unrelated stale formatting commits dropped) and squashed the two bar-chart commits into one.

The bar chart became data-export per discussion — much more useful for the paper since the visuals will get iterated on.

## How Has This Been Tested?

- `make type` and `make format` clean.
- `compare_intruder_scores.py` ran successfully against all 12 runs in the config, produced `data.json` (5.8MB) + `README.md`.
- Intruder eval rework was previously run end-to-end on the cluster (this is what the comparison data is from).

### Review hints

- Start with `spd/harvest/intruder.py` — that's the meatiest change. The streaming/lightweight-DB refactor is the architectural point; the prompt/JSON/reasoning changes are independent.
- `spd/autointerp/providers.py` — sanity-check the per-provider rate limit defaults match what you actually observe.
- `spd/harvest/db.py` — note the legacy-fallback in `get_component_densities`; flag if you want it removed.
- `spd/harvest/scripts/compare_intruder_scores.py` — pure data export, ~190 lines.
- Sample of the produced `README.md`/`data.json` lives at `SPD_OUT_DIR/intruder_comparison/` if you want to eyeball.

- `run_intruder_scoring`'s signature changed (`db` instead of `components` + `score_db`). Only caller is `spd/harvest/scripts/run_intruder.py`, which is updated.
- Per-provider rate-limit fields moved. Any existing YAML configs setting `max_concurrent` / `max_requests_per_minute` at the top level (autointerp/graph_interp/intruder configs) will need to move them under `llm:`.
- Old harvest DBs without `n_activation_examples` column still work via the `_has_column` fallback.


ocg-goodfire:
@claude pretty sure this is stale, I think a version of this made it in already. Can you check pls

ocg-goodfire:
@claude thank you

PR #481, opened 2026-05-04
Rename spd package to param_decomp

danbraunai-goodfire:
## Description

Renames the Python package and project-facing identifiers from `spd` to `param_decomp` / `PD` across the repo. This includes imports, module and file names, CLI entry points, app paths, configs, docs, tests, run metadata, environment variables, output directories, wandb defaults, and PD-facing method labels.

## Related Issue

N/A

## Motivation and Context

The repo now uses Param Decomp as the project/package name. This PR makes the codebase consistent with that name while preserving compatibility for existing stored runs, wandb artifacts, and local run caches.

Compatibility is handled explicitly:
- Historical wandb references such as `wandb:goodfire/spd/runs/...`, `goodfire/spd-pre-Sep-2025`, `goodfire/spd-tms`, and `goodfire/spd-induction-heads` are preserved.
- Old YAML configs that serialize `pretrained_model_class: spd.X` are migrated at load time to `param_decomp.X`.
- Bare run ID parsing accepts both legacy `s-...` IDs and new `p-...` IDs.
- Existing cached runs under the old layout continue to resolve.

Method labels used by autointerp prompts now use `pd` for Param Decomp. `clt` and `transcoder` remain as separate decomposition method labels for those adapter paths.

## Implementation Notes

- Moves the package from `spd/` to `param_decomp/` and updates imports throughout the backend, frontend app integration, scripts, experiments, metrics, clustering, harvest, autointerp, graph interp, and tests.
- Renames project-facing SPD types/modules to Param Decomp names, including run info, adapter/harvest modules, CLI scripts, and type files.
- Replaces `SPD_*` environment variables with `PARAM_DECOMP_*` equivalents.
- Changes new wandb runs to use the `goodfire/param-decomp` project.
- Changes the decomposition run type from `spd` / `s-...` to `param_decomp` / `p-...` for new runs, while retaining legacy parsing/loading.
- Updates Slurm naming from `spd` to `pd` where it is a project-facing prefix.

## Non-cosmetic changes

The PR is overwhelmingly mechanical rename + import updates. The behavioral changes (anything that is not a pure identifier rename) are:

- **WandB-loading bug fix in `ParamDecompAdapter`.** The adapter previously hardcoded `goodfire/spd/runs/{run_id}`, so any run trained under `goodfire/param-decomp` could not be loaded by `pd-harvest`, `pd-autointerp`, `pd-graph-interp`, or intruder eval. The adapter now takes the full `wandb_path`, and `adapter_from_id` always recovers it from the harvest DB (the `s-`-prefix short-circuit is removed).
- **Run ID format.** New runs use `RunType="param_decomp"` and a `p-` prefix. `_BARE_RUN_ID_RE` is widened so legacy `s-...` bare IDs still parse.
- **Default wandb project.** `DEFAULT_PROJECT_NAME` is now `"param-decomp"`; new runs land in `goodfire/param-decomp` instead of `goodfire/spd`.
- **Env-var compatibility removed.** The `SPD_OUT_DIR` env-var fallback is dropped entirely; `SPD_APP_DEFAULT_RUN`, `SPD_INVESTIGATION_DIR`, and `SPD_APP_DB_PATH` are renamed to `PARAM_DECOMP_*` with no aliases.
- **`WORKSPACE_TEMPLATES` cleared and `create_workspace_view` made nullable.** The old placeholder URLs were causing `pd-run` to crash in `Workspace.from_url(...PLACEHOLDER...)`. The dict is now empty, and `create_workspace_view` returns `None` for experiments without a template, so the view-creation step is skipped cleanly until templates are recreated in the new project.
- **Pydantic load-time migrations** for old YAMLs and stored harvest configs:
  - `pretrained_model_class: spd.X` → `param_decomp.X`
  - `Config.init_spd_checkpoint` → `init_pd_checkpoint`
  - `DatasetAttributionConfig.spd_run_wandb_path` → `wandb_path`
  - `ClusterMappingFile.spd_run` → `pd_run`
  - `HarvestConfig.method_config.type: "SPDHarvestConfig"` → `"ParamDecompHarvestConfig"`. Without this, every pre-rename harvest stored on NFS is unreadable: pydantic's discriminator rejects the legacy literal before the variant validator runs, so `pd-intruder`, `pd-autointerp`, and `pd-graph-interp` crash on import of the harvest config row.
- **Autointerp `activation_threshold` lookup fixed.** `autointerp/interpret.py` was reading `harvest_config["activation_threshold"]` at the top level, but the field has lived under `method_config` since the harvest schema generalization (`5a8b598f`). The lookup now goes through `method_config["activation_threshold"]`. (Pre-existing bug; surfaced because `pd-autointerp` is now exercised end-to-end in the smoke check.)
- **Decomposition method label changed from `param_decomp` to `pd`.** The `DecompositionMethod` `Literal` is `"pd" | "clt" | "transcoder"`. This propagates into autointerp prompt text (`DECOMPOSITION_DESCRIPTIONS`, `rich_examples`), `ParamDecompAdapter.model_metadata`, and the correlations router.
- **MCP server identity for the investigate agent.** The MCP server name in the generated config and the `--allowedTools` / `permissions.allow` glob both move from `spd` / `mcp__spd__*` to `param_decomp` / `mcp__param_decomp__*`. These two strings must agree for tool calls to resolve.
- **`wandb.run.log_code` root path.** Switched from `REPO_ROOT / "spd"` to `REPO_ROOT / "param_decomp"`; without this, code logging would silently log nothing after the package move.
- **Slurm job prefix.** `slurm_job_name_prefix` in `pipeline_config.yaml` goes from `spd` to `pd`, so submitted jobs change name.
- **Frontend canonical-run registry pruned.** Four stale entries removed from `app/frontend/src/lib/registry.ts`; the remaining `Jose` entry is annotated as the `VPD paper run`. Visible in the app's run picker.
- **New test** `tests/autointerp/test_decomposition_method.py` pins `DecompositionMethod` to `{"pd", "clt", "transcoder"}` and asserts no `spd` / `param_decomp` key leaks back in.

## How Has This Been Tested?

- AI helper ran all (or at least most) scripts in the repo, checked that they ran to completion and that the outputs were in the expected dir.

Yes. Code importing from `spd` must update to `param_decomp`, and users must replace `SPD_OUT_DIR`, `SPD_APP_DEFAULT_RUN`, `SPD_INVESTIGATION_DIR`, and `SPD_APP_DB_PATH` with the corresponding `PARAM_DECOMP_*` environment variables. Stored wandb artifacts and existing on-disk run caches remain supported.


oclivegriffin (approved):
Looks sensible. tbh I though I don't think I'm gonna catch anything that's been missed. Seems like Claude did a good job validating

ocg-goodfire (approved):
Looks sensible. tbh I though I don't think I'm gonna catch anything that's been missed. Seems like Claude did a good job validating

PR #483, opened 2026-05-06
Rust accel scoring

patrickdugan:
## Description
  Adds an optional Rust/PyO3 acceleration backend for rank-one component scoring.

  The default behavior remains the existing Python path. Rust acceleration is opt-in via `backend="rust"` or
  `PARAM_DECOMP_ACCEL=rust`, with `backend="auto"` available for explicit fallback behavior. The PR also adds support
  for separate `metric_reference_logits`, so ablation logits can be computed from one reference tensor while metrics are
  evaluated against another.

  Included changes:
  - New `param_decomp.accel.scoring` Python wrapper.
  - New `param_decomp_accel` Rust/PyO3 crate.
  - Rust implementation of rank-one linear component ablation scoring.
  - Unit tests covering Python/Rust parity, backend selection, validation, and metric-reference behavior.
  - Benchmark script for comparing Python vs Rust scoring.
  - README/docs updates showing opt-in usage.

  ## Related Issue
  N/A

  ## Motivation and Context
  Component scoring is a bottleneck in VPD-style decomposition loops, especially when repeatedly evaluating many rank-
  one ablations across bootstrap samples. The existing Python implementation is correct but can spend most of runtime in
  the scoring hot loop.

  This change creates a narrow acceleration seam: orchestration, decomposition, and model logic stay in Python, while
  the tight numeric scoring loop can be moved to Rust when requested. Keeping Rust mode opt-in preserves compatibility
  and makes the change viable for incremental adoption or PR review.

  ## How Has This Been Tested?
  Tested locally on Windows.

  Unit tests:
  ```powershell
  python -m pytest tests\accel\test_scoring.py -q

Result:

  7 passed

  Rust tests:

  cargo test --manifest-path crates\param_decomp_accel\Cargo.toml

  Result: passed.

  Local benchmark/smoke testing was also run against the accelerated scoring path after rebuilding and reinstalling the
  PyO3 wheel with maturin. The Rust backend matched Python scoring behavior and produced a substantial speedup in the
  target rank-one scoring loop.

  No.

  The default backend remains Python. Rust acceleration is only used when explicitly requested with backend="rust",
  backend="auto", or PARAM_DECOMP_ACCEL=rust. Existing callers should continue to behave the same unless they opt into
  the new backend.


danbraunai-goodfire:
Written by AI:

Thanks for putting this together, but unfortunately it's a bit too cumbersome for us to review and accept PRs from external contributors, so we won't be merging this. Closing without prejudice — the work isn't wasted, just not something we can take on through this channel.

In case any of it is useful to you, a few observations from a read-through:

**Whether this is actually a hot loop here.** The motivation cites "VPD-style decomposition loops" and "bootstrap component scoring", but there's no caller for `score_rank_one_linear_components` in `param_decomp/`, and the metric names the kernel emits (`ablated_behavior_mse`, `ablated_task_loss`, `ablated_accuracy`, `ablated_original_pred_match_rate`) don't appear anywhere in the codebase. The actual training hot loop in `param_decomp/run_param_decomp.py:248` is GPU PyTorch (DDP, bf16 autocast) — forward/backward through the component model with stochastic masks — and doesn't compute single-component rank-1 ablations on CPU activations.

Separately, the Python "baseline" being benchmarked loops over components in Python and does four separate numpy reductions per component. A batched rewrite (`scales = inputs @ V.T` once, then a single vectorised expression) would likely close most of the gap without any Rust. And if scoring genuinely is large enough to matter, GPU `torch.einsum`/`bmm` would be a more natural fit than CPU+Rayon given the rest of the stack already runs on GPU.

**Supply-chain surface.** Compiled native extensions run with the Python process's full privileges, and this adds a new one plus ~25 transitive Rust crates. The Rust source itself is straightforward numerics (no `unsafe`, no `build.rs`, no network/FS access, all deps from crates.io with normal checksums), but adopting it would mean owning that surface. The published `pyproject.toml` also claims the PyPI name `param-decomp-accel`, which appears unowned — worth claiming if you continue with this in a fork.

**\"Storyworld\" reference.** `docs/rust_accel_scoring.md` mentions wiring this to \"Goodfire/Storyworld bootstrap scoring jobs\", which isn't part of this repo, and the README change reads \"This fork includes…\" — so this seems intended for a downstream pipeline rather than upstream `param-decomp` proper. That's probably the right home for it.

A few smaller things in case they help downstream:

- `param_decomp/accel/scoring.py` uses `from __future__ import annotations`, which the project `CLAUDE.md` explicitly forbids.
- Rust `validate_shapes` only checks the first/last entries of `slice_offsets`. A negative or out-of-range middle value will panic in `slice_indices[start..end]`. The Rust path also doesn't re-validate that labels are in `[0, out_dim)`, so direct callers of the extension can panic in `cross_entropy_row`.
- `_resolve_backend` honors `PARAM_DECOMP_ACCEL` from the env, which is implicit activation; explicit `backend=` would be more in keeping with the project's style.

Thanks again for the interest in the project.

patrickdugan:
The feedback was very detailed, it's true that this acceleration was
designed for a different CPU/RAM based stack focusing on rapid
iteration on very small models, like 7M TRMs.

Thanks!

PR #486, opened 2026-05-08
The sexy refactor

danbraunai-goodfire:
## Description

Split the codebase into two flat-layout distributions and reduce the core training surface to a single entrypoint:

- **`param-decomp`** (core): the reusable, publishable library — `optimize()`, `ComponentModel`, configs, loss metrics, and the `RunSink` protocol. Treat as a stable API.
- **`param-decomp-lab`** (team tooling): experiment scripts, `pd-*` CLIs, eval metrics, postprocess pipelines, the app, infra, and the concrete `RunSink` (local files + wandb + rank-aware no-op). Churns freely; depends on core.

The trainer is now a pure function with no driver class, no YAML parsing, no wandb init:

```python
from param_decomp.optimize import optimize

optimize(
    target_model,           # nn.Module, frozen by optimize()
    train_loader,           # DataLoader
    run_batch,              # RunBatch callable — one batch through the wrapped model
    reconstruction_loss,    # ReconstructionLoss callable — (loss_sum, n_elements)
    pd_config,              # PDConfig    — algorithm: CI fn, loss metrics, optimizers, targets, …
    runtime_config,         # RuntimeConfig — device, autocast, DDP
    sink,                   # RunSink     — log / console / checkpoint protocol
    cadence,                # Cadence     — train-log + save-every predicates
    eval_loop=None,         # EvalLoop | None — optional eval bundle
)
```

Each experiment's `run.py` is now a plain composition root: parse YAML into an `ExperimentConfig`, build the target / loaders / metrics / sink, call `optimize()`.

## Related Issue

N/A

## Motivation and Context

The previous architecture had core code reaching into orchestration, sweep handling, and per-experiment driver classes. Splitting core/lab makes the trainer reusable as a library (notebook callers can `from param_decomp.optimize import optimize` with no `pd-*` baggage) and gives the lab room to grow without churning the core.

## How Has This Been Tested?

- `make check` clean (basedpyright 0 errors, ruff lint+format clean).
- `make test` 98/98 pass.
- All in-repo YAMLs (tms / resid_mlp / lm) validate end-to-end via `<Experiment>ExperimentConfig.from_file(...)` with `extra="forbid"`.

Yes — consistent with the repo's no-legacy-shim policy:

- YAML schema: `pd:` / `runtime:` / `cadence:` / `target:` / `data:` / `eval:` (was `pd:` / `runtime:` / `target:` / `data:` / `logging:`). Frequency keys renamed `*_freq` → `*_every`. All in-repo configs are updated.
- Public API: import from definition modules, not package roots. The `pd-experiment` driver CLI is removed — use `pd-tms` / `pd-resid-mlp` / `pd-lm` (or call a `run.py` directly).
- Saved runs: `experiment_manifest.yaml` → `run_meta.yaml` with a new shape. Old runs need manual migration or retrain.


param_decomp/adapters/param_decomp.py
```diff
@@ -49,28 +54,39 @@ def layer_activation_sizes(self) -> list[tuple[str, int]]:
...
+        assert isinstance(exp, LMExperimentConfig), f"No tokenizer for kind={exp.kind!r}"
+        return exp.data.tokenizer_name
 
     @property
     @override
     def model_metadata(self) -> ModelMetadata:
-        cfg = self.pd_run_info.config
-        task_cfg = runtime_cast(LMTaskConfig, cfg.task_config)
+        exp = self.experiment_config
+        assert isinstance(exp, LMExperimentConfig), (
```
ocg-goodfire: this assertion is made twice. can we encode it in the types?

param_decomp/adapters/param_decomp.py
```diff
@@ -49,28 +54,39 @@ def layer_activation_sizes(self) -> list[tuple[str, int]]:
...
                 for path in self.component_model.target_module_paths
             },
-            seq_len=task_cfg.max_seq_len,
+            seq_len=exp.data.max_seq_len,
             decomposition_method="pd",
         )
+
+
+# Back-compat alias. Removed in step 9.
+ParamDecompAdapter = PDAdapter
```
ocg-goodfire: meant to still be here?

param_decomp/models/batch_and_loss_fns.py
```diff
@@ -30,6 +31,50 @@ class ReconstructionLoss(Protocol):
...
+    and its return value is what `run_batch` and every downstream loss/metric sees.
+    Override the default `move_batch_to_device` if your dataloader yields a structure
+    where only some fields belong on device (e.g. a dict with large unused keys).
+    """
+
+    model: nn.Module
+    run_batch: RunBatch
+    reconstruction_loss: ReconstructionLoss
+    tied_weights: list[tuple[str, str]] | None = None
+    name: str = "custom"
```
ocg-goodfire: should this be `kind` (seems to be the same as Driver.kind?)

param_decomp/run_param_decomp.py
```diff
@@ -425,26 +428,26 @@ def create_pgd_data_iter() -> Iterator[Any]:
...
     run_id: str | None = None,
-    launch_id: str | None = None,
-    evals_id: str | None = None,
     sweep_params: dict[str, Any] | None = None,
-    target_model_train_config: BaseConfig | None = None,
-    tied_weights: list[tuple[str, str]] | None = None,
-) -> None:
-    """Run a full PD experiment: setup, optimize, cleanup.
+    manifest: ExperimentManifest | None = None,
+    artifacts: Sequence[RunArtifact] = (),
```
ocg-goodfire: why are we using this type signature instead of just a list? tuples seem weird

ocg-goodfire (commented):
a few small comments I'll leave now while I read in cursor

ocg-goodfire:
see https://github.com/goodfire-ai/param-decomp/pull/492

param_decomp_lab/experiments/lm/data.py
```diff
@@ -0,0 +1,313 @@
...
+from torch import Tensor
+from torch.utils.data import DataLoader, DistributedSampler
+from transformers import AutoTokenizer, PreTrainedTokenizer
+
+from param_decomp.base_config import BaseConfig
+from param_decomp.log import logger
+from param_decomp.utils.distributed_utils import DistributedState
+
+
+class LMDataConfig(BaseConfig):
```
ocg-goodfire, line 20: wondering whether this should own: `buffer_size` (higher confidence) and `batch_size` (lower)

param_decomp/experiments/lm/data.py
```diff
@@ -0,0 +1,313 @@
...
+    is_tokenized: bool
+    hf_tokenizer_path: str
+    streaming: bool
+    split: str
+    n_ctx: int
+    """Must be model n_ctx + 1 to provide room for next-token label indexing."""
+    seed: int | None = None
+    column_name: str
+    """Dataset column containing text or token ids."""
+    shuffle_each_epoch: bool = True
```
ocg-goodfire: maybe out of scope but these 2's coexistence is weird to me. Looks like it's cos we do things differently in spd/pretrain, but why?

param_decomp_lab/experiments/lm/data.py
```diff
@@ -0,0 +1,313 @@
...
+    is_tokenized: bool,
+    streaming: bool,
+    column_name: str,
+    batch_size: int,
+    buffer_size: int,
+    seed: int,
+    shuffle_each_epoch: bool = True,
+    dist_state: DistributedState | None = None,
+    collate_fn: Callable[..., Any] | None = None,
+) -> tuple[DataLoader[Any], PreTrainedTokenizer]:
```
ocg-goodfire, line 153: pending (https://github.com/goodfire-ai/param-decomp/pull/486/changes#r3235414936), could this just take a `LMDataConfig` instead of many of these args?

ocg-goodfire (commented):
I wish github would just post comments as you did them lol. Still reviewing

param_decomp/load.py
```diff
@@ -0,0 +1,41 @@
...
+
+from pathlib import Path
+
+from param_decomp.models.batch_and_loss_fns import PDTarget
+from param_decomp.models.component_model import ComponentModel, PDRunInfo
+from param_decomp.param_decomp_types import ModelPath
+from param_decomp.utils.general_utils import fetch_latest_local_checkpoint
+
+
+def load_pd(
```
ocg-goodfire: should this be called `load_component_model`? just trying to get all the nouns super clear in my head

param_decomp/adapters/param_decomp.py
danbraunai-goodfire: Fixed

param_decomp/models/batch_and_loss_fns.py
danbraunai-goodfire: Fixed

param_decomp/run_param_decomp.py
danbraunai-goodfire: I think AI did this because it lets you set an immutable default of (). If it was a list, you have to do `artifacts: list[RunArtifact] = field(default_factory=list)` and:
```
artifacts: list[RunArtifact] | None = None,
...
artifacts = [] if artifacts is None else artifacts
```
I'm actually on board with this little code-saving trick.

param_decomp/experiments/lm/data.py
danbraunai-goodfire: Good call. I combined these. Left with:

param_decomp_lab/experiments/lm/data.py
danbraunai-goodfire: after the change, we're left with:
```
def create_lm_data_loader(
    cfg: LMDataConfig,
    *,
    split: str,
    batch_size: int,
    seed: int,
    dist_state: DistributedState | None = None,
    collate_fn: Callable[..., Any] | None = None,
) -> tuple[DataLoader[Any], PreTrainedTokenizer]:
```
Those arguments do get passed different values for the same config.

param_decomp_lab/experiments/lm/data.py
danbraunai-goodfire: it now owns buffer_size but not batch_size. I think the latter is too much of a PD hyperparameter to house elsewhere.

param_decomp/adapters/param_decomp.py
danbraunai-goodfire: Removed along with all other backward compatibility handling

ocg-goodfire:
@claude Can you do a quick audit of the main Nouns is the codebase, especially those touched and added here? As a motivating example, I'm wondering whether `ParamDecompHarvestConfig` and `PDAdapter` should actually be `ComponentModelHarvestConfig` and `ComponentModelAdapter`. Basically - are our Nouns well "normalised" and are we using them right everywhere

param_decomp/load.py
danbraunai-goodfire: yep, updated (and moved to saved_run.py to avoid this single function file.

PR #487, opened 2026-05-11
Refactor ComponentModel to fused DecomposedSite tree (PR 1: bones, core library only)

ocg-goodfire:
## Summary

PR 1 of the rollout in [`fsdp_implementation_plan.md`](./fsdp_implementation_plan.md). Empirical case for going straight to FSDP is in [`investigation_results.md`](./investigation_results.md) — ZeRO-1+ckpt OOMs at 4B target on a 150 GB H200, so we need full parameter sharding. This PR is the architectural prerequisite: stop using sibling `nn.ModuleDict` + forward hooks, install fused `DecomposedSite` modules in place in the target tree instead.

**Draft.** Limited to the **core library** — `param_decomp/models/`, `run_param_decomp.py`, `utils/logging_utils.py`, `metrics/`, plus minimal type-only fixes to plotting helpers. App / harvest / dataset_attributions / clustering / editing / autointerp / adapters caller updates are deliberately deferred.

## What's here

- **New file** `param_decomp/models/decomposed_module.py` — `DecomposedLinear` and `DecomposedEmbedding`. Each wraps a frozen target submodule (`self.linear` / `self.embedding`), owns its own trainable `V` and `U`, and dispatches between the wrapped target's forward (when no mask is bound) and the legacy decomposed-forward math (when bound via `site.bind(mask_info, cache, cache_type)`).
- **New file** `param_decomp/models/mask_info.py` — `ComponentsMaskInfo`, `WeightDeltaAndMask`, `make_mask_infos` moved out of `components.py` to break the import cycle between `components.py` and `decomposed_module.py`. `components.py` re-exports them for back-compat.
- **`ComponentModel` rewritten** — no more sibling `_components` ModuleDict. `install_decomposed_sites` replaces target submodules in place; `forward()` uses an `ExitStack` of `site.bind(...)` context managers instead of `_attach_forward_hooks`. The hook-based `_components_and_cache_hook` is gone.
- **CI fn wrappers** (`LayerwiseCiFnWrapper`, `GlobalCiFnWrapper`) take a `sites` dict typed against a `SiteWithComponentActs` Protocol.

## What this *unblocks* (PR 2 in the plan)

The forward hook crossing FSDP-unit boundaries was the load-bearing reason §5b of the report flagged "FSDP is not a drop-in replacement." With the fused-site shape, FSDP's auto-wrap policy can target `(DecomposedLinear, TransformerBlock)` and components live in the same unit as the target weight they replace. The actual FSDP wrap lands in PR 2.

## Tests

`make test` passes (268 passed, 13 skipped). The 13 skipped: `tests/test_wandb_run_loading.py::*` (pre-existing wandb-dependent), `test_replacing_identity` (intentionally skipped — see below).

Specific fixes inside this PR:
- `get_grad_norms_dict` and the optimizer param-collection loop in `run_param_decomp.py` filter on `requires_grad` so the frozen wrapped target's weight isn't pulled into the trainable set.
- `metrics/attn_patterns_recon_loss.py` switched from `components[name](x, mask=…)` to the new `components[name].apply_decomposed(x, mask=…)` accessor.
- Various tests use the new `component_weight` / `target_weight` accessor names where they previously relied on `components[name].weight` meaning V@U.

## Known limitations / not in this PR

- **Identity-shim decomposition** (`identity_module_info` + `insert_identity_operations_`) is dropped. Tests using it pass `identity_module_info=None` or are skipped with a pointer to `fsdp_implementation_plan.md`. The feature wasn't load-bearing for Jose/Thomas; we can revive it if a real consumer surfaces.
- **No migration script yet.** Existing Jose/Thomas checkpoints have the old sibling-ModuleDict state-dict layout (`_components.h-0-mlp-c_fc.V` etc.); under the new layout those keys are `target_model.h.0.mlp.c_fc.V`. A one-shot rename script is on the PR 1 follow-up list.
- **No loss-trajectory equivalence test yet.** A 50-step DDP run on Jose against `main` should give matching loss within numerical noise — TODO before this PR comes out of draft.
- **App / harvest / dataset_attributions / clustering / editing / autointerp / adapters** still reference `model.components[name]` as if it were a legacy `LinearComponents`. They mostly work via duck typing (V, U are still there) but their type annotations and any `.weight` / `.component_weight` accesses need a sweep.
- **No FSDP wrap policy** — that's PR 2.

## Test plan
- [x] `make test` passes locally (DDP/single-GPU)
- [ ] Loss-trajectory equivalence test on Jose under DDP (50 steps, synthetic data) — TODO
- [ ] Migration script for old checkpoints — TODO
- [ ] App / harvest / dataset_attributions / clustering / editing / autointerp / adapters callers — TODO (follow-up PRs)

### DIFF
--- a/param_decomp/models/components.py
+++ b/param_decomp/models/components.py
@@ -298,6 +298,7 @@ class GlobalSharedTransformerCiFn(nn.Module):
         mlp_hidden_dims: list[int] | None = None,
         max_len: int = 2048,
         rope_base: float = 10000.0,
+        gradient_checkpointing: bool = False,
     ):
         super().__init__()
 
@@ -307,6 +308,7 @@ class GlobalSharedTransformerCiFn(nn.Module):
         self.d_model = d_model
         self.n_transformer_layers = n_layers
         self.n_heads = n_heads
+        self.gradient_checkpointing = gradient_checkpointing
 
         if mlp_hidden_dims is None:
             mlp_hidden_dims = [4 * d_model]
@@ -349,8 +351,14 @@ class GlobalSharedTransformerCiFn(nn.Module):
             added_seq_dim = True
 
         x = projected
-        for block in self._blocks:
-            x = block(x)
+        if self.gradient_checkpointing and self.training:
+            from torch.utils.checkpoint import checkpoint
+
+            for block in self._blocks:
+                x = checkpoint(block, x, use_reentrant=False)
+        else:
+            for block in self._blocks:
+                x = block(x)
 
         output = self._output_head(x)
 
--- a/param_decomp/models/components.py
+++ b/param_decomp/models/components.py
@@ -1,13 +1,26 @@
 from abc import ABC, abstractmethod
+from collections.abc import Mapping
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal, override
+from typing import TYPE_CHECKING, Protocol, override
 
 import einops
 import torch
 import torch.nn.functional as F
-from jaxtyping import Bool, Float, Int
+from jaxtyping import Float, Int
 from torch import Tensor, nn
 
+from param_decomp.models.mask_info import (
+    ComponentsMaskInfo as ComponentsMaskInfo,  # re-export for backward compat
+)
+from param_decomp.models.mask_info import (
+    RoutingMasks as RoutingMasks,  # re-export for backward compat
+)
+from param_decomp.models.mask_info import (
+    WeightDeltaAndMask as WeightDeltaAndMask,  # re-export for backward compat
+)
+from param_decomp.models.mask_info import (
+    make_mask_infos as make_mask_infos,  # re-export for backward compat
+)
 from param_decomp.utils.module_utils import _NonlinearityType, init_param_
 
 if TYPE_CHECKING:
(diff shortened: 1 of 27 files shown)


param_decomp/models/components.py
```diff
@@ -479,6 +505,48 @@ def forward(
...
+    def forward_with_target_weight(
+        self,
+        x: Float[Tensor, "... d_in"],
+        target_weight: Float[Tensor, "d_out d_in"],
+        mask: Float[Tensor, "... C"] | None = None,
+        weight_delta_mask: Float[Tensor, "..."] | None = None,
+        component_acts_cache: dict[str, Float[Tensor, "... C"]] | None = None,
+    ) -> Float[Tensor, "... d_out"]:
+        """Equivalent to `forward` with `weight_delta_and_mask=(target_weight - V@U, weight_delta_mask)`,
+        but never materializes the (d_out, d_in) delta tensor.
```
danbraunai-goodfire, line 517: This function name is confusing. This is a masked forward pass. Shouldn't have target_weight in the name. maybe `forward_with_mask` is fine (possible I'm missing something).

danbraunai-goodfire (commented):
Not sure how done you are with this. I'll look deeper if you're happy with it

ocg-goodfire:
Superseded by #488, which now targets main directly and includes everything from this PR plus the FSDP2 wrap.

PR #489, opened 2026-05-12
Refactor loss/eval metric configs to named-field containers

danbraunai-goodfire:
## Description

Replace `PDConfig.loss_metric_configs` and `PDConfig.eval_metric_configs` (lists of discriminated-union configs) with `loss_metrics: LossMetricsConfig` and `eval_metrics: EvalMetricsConfig`. Each container exposes one nullable field per available metric; setting a field selects the metric, leaving it `None` omits it. The field name is the discriminator — no more `classname:` boilerplate in YAML.

Before:
```yaml
loss_metric_configs:
- classname: ImportanceMinimalityLoss
  coeff: 3e-3
  pnorm: 1.0
  beta: 0
- classname: StochasticReconLoss
  coeff: 1.0
```

After:
```yaml
loss_metrics:
  importance_minimality:
    coeff: 3e-3
    pnorm: 1.0
    beta: 0
  stochastic_recon:
    coeff: 1.0
```

Other changes that fall out:
- `get_unique_metric_configs` deleted: effective eval set is just `loss_metrics.active() + eval_metrics.active()`. Loss metrics still auto-flow into eval.
- A validator rejects the same metric set on both sides (replaces the previous silent eval-wins warning).
- Persistent-PGD and multibatch-PGD configs are now read directly from named fields rather than `isinstance`-filtered from a list.
- `_DISCRIMINATED_LIST_FIELDS` and ~120 lines of discriminator special-casing removed from `apply_nested_updates` and the sweep-params validators in `utils/run_utils.py`.
- `flatten_metric_configs`, `_parse_metric_config_key`, and `generate_wandb_run_name` rewritten to use snake_case field-name prefixes (`loss.ImpMin.coeff` instead of `loss_metric_configs.ImportanceMinimalityLoss.coeff`).
- All 12 built-in experiment YAMLs and the sweep-params YAMLs migrated to the new shape.

## Motivation and Context

The list-of-discriminated-unions shape had several costs:
- The `classname` discriminator was also an informal uniqueness key (you almost never want two `FaithfulnessLoss` configs), but uniqueness was only enforced at runtime via a warning.
- Static access required `isinstance` scans over the list at every call site.
- The sweep / grid / wandb-flattening utilities special-cased these two fields heavily.
- YAML carried `classname:` boilerplate, and eval-only entries awkwardly set `coeff: null` for classes shared between loss and eval.

The named-field shape removes all of that, gives `PDConfig.loss_metrics.faithfulness` a real static type, and lets the existing nested-dict sweep/grid machinery work without any discriminator-aware code path.

## How Has This Been Tested?

- [x] `make check` (basedpyright + ruff) — clean
- [x] `make test` — 462 passed, 5 skipped (slow tests excluded; same as before)
- [x] Each built-in experiment YAML loads via its driver's `config_model.from_file(...)` and round-trips through `LossMetricsConfig.active()` / `EvalMetricsConfig.active()`
- [x] `tests/scripts_run/test_grid_search.py` rewritten to exercise the simplified sweep path and passes
- [x] `tests/test_distributed.py`, `tests/test_gpt2.py`, `tests/test_component_model.py`, `tests/test_tms.py`, `tests/test_resid_mlp.py`, `tests/test_experiment_manifest.py` all updated to construct the new shape and pass

**Yes** — but limited and intentional, with no migration shim (per design discussion):

- Old `PDConfig` instantiations with `loss_metric_configs=...` / `eval_metric_configs=...` raise an assertion at validation time with a clear migration message pointing at the new field name.
- Saved manifests (`experiment_manifest.yaml`) on disk or in wandb that embed the old list format will fail to reload via `PDRunInfo.from_path`. Historical runs that need reloading must be hand-edited.
- Sweep YAMLs using the old `loss_metric_configs: - classname: X` form need to be converted to the new nested-dict form.

Field-name mapping is `<ClassName>` → snake_case minus the `Loss` suffix (e.g. `ImportanceMinimalityLoss` → `importance_minimality`, `StochasticReconSubsetLoss` → `stochastic_recon_subset`, `CIHistograms` → `ci_histograms`).


param_decomp/utils/run_utils.py
ocg-goodfire, line 1: lol love it

param_decomp/utils/wandb_utils.py
```diff
@@ -222,38 +214,32 @@ def parse_wandb_run_path(input_path: str) -> tuple[str, str, str]:
...
-
-            classname = config_item["classname"]
-            assert isinstance(classname, str), f"{config_list_name} should have a classname"
-            short_name = METRIC_CONFIG_SHORT_NAMES[classname]
-
-            for key, value in config_item.items():
+        container = config_dict[container_name]
+        assert isinstance(container, dict), f"{container_name} should be a dict"
+
+        prefix = container_name.split("_")[0]  # "loss" or "eval"
```
ocg-goodfire, line 233:
```suggestion
        prefix = { "loss_metrics": "loss", "eval_metrics": "eval" }[container_name]
```

param_decomp/configs.py
```diff
@@ -195,7 +195,7 @@ class ModulePatternInfoConfig(BaseConfig):
 class LossMetricConfig(BaseConfig):
     coeff: float | None = Field(
         default=None,
-        description="Loss coefficient. Used when metric is in loss_metric_configs.",
+        description="Loss coefficient. Required when set under `loss_metrics`; ignored under `eval_metrics`.",
     )
```
ocg-goodfire, line 153: now that we never rely on this interface it can be deleted

param_decomp/configs.py
```diff
@@ -599,6 +599,78 @@ class UVPlotsConfig(BaseConfig):
...
+    pgd_recon_subset: PGDReconSubsetLossConfig | None = None
+    pgd_recon_layerwise: PGDReconLayerwiseLossConfig | None = None
+    persistent_pgd_recon: PersistentPGDReconLossConfig | None = None
+    persistent_pgd_recon_subset: PersistentPGDReconSubsetLossConfig | None = None
+
+    def active(self) -> list[LossMetricConfigType]:
+        return [v for _, v in self if v is not None]
+
+
+class EvalMetricsConfig(BaseConfig):
```
ocg-goodfire: should/could this inherit `LossMetricsConfig`?

ocg-goodfire (approved):
Nice. this is cleaner for sure

param_decomp/configs.py
danbraunai-goodfire: We do use it though, the metrics we have that can be losses are subclasses of it because they have a coeff field.

param_decomp/configs.py
danbraunai-goodfire: Done

param_decomp/configs.py
danbraunai-goodfire: I don't get this. We do use this interface?

param_decomp/utils/wandb_utils.py
danbraunai-goodfire: done

param_decomp/configs.py
ocg-goodfire: ah ok fair. I was fixating on the fact that we never rely on it, as in, no code ever expects, or defines behaviour in terms of a LossMetricConfig (eg: it's not references outside the file).

ocg-goodfire: the thing that set off my instincts were that it doesn't actually guarantee any behaviour cos the field is nullable anyway. like, to a consumer, knowing a thing is a `LossMetricConfig` is useless

ocg-goodfire: ideally we'd only use if for loss metrics and not inherit in the eval case, but that's a lot of boilerplate. @claude any suggestions for a nice pattern here? runtime inheritance perhaps? like I want this:

(very loosely)
```python
class LossConfigs:
    recon: inhereit(Recon, LossMetricConfig)

class EvalConfigs:
    recon: Recon
```

danbraunai-goodfire: @ocg-goodfire I'm not a fan of either of these solutions. Leaning towards merge as is.

danbraunai-goodfire: Actually, maybe the LossWrapper option is fine.

ocg-goodfire: yea I think just merging as-is is fine. this is overcomplicated probably

PR #492, opened 2026-05-13
Simplify ExperimentConfig, kind, and dataset seed

ocg-goodfire:
## Description

Three small follow-on simplifications on top of the modular-driver refactor.

**1. Strip `ExperimentConfig` to one field.**
- Removed dead `metadata: dict[str, Any] | None` and redundant `kind: str` from the base.
- Concrete configs (`TMSExperimentConfig`, `LMExperimentConfig`, `ResidMLPExperimentConfig`) no longer redeclare `kind`.
- `kind` now has exactly two homes: the driver class (`ClassVar[str]`, source of truth) and the manifest (`kind: str`, durable snapshot populated from `driver.kind`).
- Downstream consumers that previously read `experiment_config.kind` (clustering, dataset attributions, adapters, app router, tests) now read `run_info.manifest.kind` instead.

**2. Collapse the LM dataset seed.**
- `LMDataConfig.dataset_seed: int | None = None` (with a three-level fallback through `pd.seed`) → `dataset_shuffle_seed: int = 0`. Single source, no nullable, no falsy default chain.
- Removed `seed: int | None` from `ExperimentDriver.build_dataloaders` and `PDRunInfo.build_dataloaders`. Seed lives in the config, not as a runtime side-input.
- `build_lm_dataloaders` no longer takes `seed` / `default_seed`.

**3. Clustering harvest uses `model_copy` instead of mutating frozen config.**
- `run_harvest.py` was assigning `pd_run.experiment_config.data.dataset_shuffle_seed = config.dataset_seed`, which raises `ValidationError: Instance is frozen` (`BaseConfig` has `frozen=True`).
- Now it `model_copy`s the `LMDataConfig` with the per-run seed and calls `build_lm_dataloaders` directly. Driver layer bypassed for this one out-of-band concern; manifest and shared state untouched.

Also: removed `from __future__ import annotations` from a handful of touched files (per project convention) and string-quoted the forward refs that needed it.

## Motivation and Context

Polishing the open-driver refactor:
- `ExperimentConfig.kind` was a phantom discriminator that no closed-world dispatch ever consumed — keeping it invited misuse.
- `ExperimentConfig.metadata` had no readers anywhere.
- The dataset-seed fallback chain was three indirections deep for what is conceptually one value.
- The clustering harvest's in-place config mutation was a latent crash.

## How Has This Been Tested?

- `pytest tests/` — all tests pass (including the round-trip manifest tests, which now source `kind` from the driver).
- `basedpyright` clean on all touched files.
- Verified the frozen-mutation crash reproduces on `main` and is fixed here.

Yes, but narrowly: saved manifests from before this PR still load (manifest shape is unchanged — `kind` was already a top-level manifest field), but **any caller reading `experiment_config.kind` on the parsed config will break**. Migration path: read `run_info.manifest.kind` instead.


param_decomp/experiments/lm/experiment.py
```diff
@@ -107,10 +106,18 @@ def load_lm_target(target_cfg: LMTargetConfig) -> PDTarget:
...
+    def kind(self) -> str:
+        return "lm"
 
+    @property
+    @override
+    def config_model(self) -> type[LMExperimentConfig]:
+        return LMExperimentConfig
+
+    @override
     def prepare(
```
ocg-goodfire, line 121: the point of this pattern is to enable definition-time typechecking. still allows others to use the protocol, but it's convenient to get errors here instead of where it's passes as an `ExperimentDriver[T]`

param_decomp/experiments/driver.py
```diff
@@ -5,30 +5,24 @@
...
 from param_decomp.base_config import BaseConfig
 from param_decomp.configs import PDConfig
+from param_decomp.models.batch_and_loss_fns import PDTarget
+from param_decomp.utils.distributed_utils import DistributedState
 
 
 class ExperimentConfig(BaseConfig):
     """Pure-data config shared by all experiment configs."""
 
-    kind: str
```
ocg-goodfire, line 29: This was duplicate state from the driver

param_decomp/experiments/driver.py
```diff
@@ -5,30 +5,24 @@
...
     pd: PDConfig
-    metadata: dict[str, Any] | None = None
```
ocg-goodfire, line 31: was unused

tests/test_lm_data.py
ocg-goodfire, line 1: these were all testing the seed behaviour which is now self-evidently correct

danbraunai-goodfire (approved):
nice

danbraunai-goodfire:
@ocg-goodfire I reverted this seed thing in bab83051. I removed the seed option from the LMDataConfig. It now lives in PDConfig, and as passed as its own argument to create_dataloader functions.

The reason is that we frequently want to sweep over a single seed argument, and have that vary the model init and dataset seed. And I think it's unlikely that in the main code we'll want to sweep over the dataset seed but not that model init seed.

PR #494, opened 2026-05-14
Store run snapshots under refs/runs/* instead of branches

ocg-goodfire:
## Description

Run snapshots are currently pushed as `snapshot/<id>` branches under `refs/heads/*`. Because every client's `git fetch` pulls all of `refs/heads/*` by default, the ~1500 snapshot branches we've accumulated end up in every clone — slowing `git pull` for no benefit (no one ever checks them out locally).

This PR moves the snapshot ref into a custom namespace, `refs/runs/snapshot/<id>`, which a default `git fetch` ignores. The commit itself (including folded-in uncommitted changes) is unchanged — only the ref location differs.

## Motivation and Context

`git pull` was dragging because we accumulate a snapshot branch per run, forever. Custom ref namespaces are the right git primitive for "server-side reachable, but not synced by default."

Trade-offs considered:
- **Tags** — would work for read but `git push -u` doesn't apply, and tag push syntax differs; would have required branching logic.
- **Custom ref namespace** — one-line change in the creator, all push/checkout syntax identical. Picked this.

## What changed

- `create_git_snapshot()` now writes `refs/runs/snapshot/<id>` via `git update-ref` on a detached worktree, then pushes that refspec. Pre-commit hooks are bypassed on the snapshot commit exactly as before.
- SLURM workspace setup now does an explicit `git fetch <repo> <ref>:<ref>` after `git clone` (since `git clone` only pulls `refs/heads/*` + tags by default), then `git checkout <ref>`.
- `snapshot_branch` renamed to `snapshot_ref` everywhere for honest naming.
- New script `scripts/migrate_snapshot_branches_to_runs_refs.sh` to move existing `snapshot/*` branches on origin to `refs/runs/snapshot/*` and delete the old ones. Dry-run by default; `--apply` to execute. Idempotent.

## How Has This Been Tested?

- Smoke-tested `create_git_snapshot()` against this repo end-to-end: ref created locally, pushed to origin, resolved by SHA, cleaned up. Pushed and then deleted `refs/runs/snapshot/smoketest-DELETEME` on origin successfully.
- Dry-ran the migration script: detected 1481 `snapshot/*` branches on origin and planned the rewrite. Idempotent path (already-migrated refs) re-checked.
- `make type` and `ruff check` pass on all touched files.

Not tested:
- Actual SLURM job. The new `git fetch <repo> <ref>:<ref>` step is the only new git operation in the SLURM path; nothing else about the job flow changes.

## Migration plan

After this PR merges:
1. Run `./scripts/migrate_snapshot_branches_to_runs_refs.sh --apply` once from a machine with push access. This is what makes everyone's local clones lighter going forward.
2. Teammates can prune their local `refs/remotes/origin/snapshot/*` with `git remote prune origin` after the next fetch.

Old saved metadata (W&B reports, manifest YAMLs) that recorded the old branch name will no longer resolve via that ref — they still reference the commit SHA, which is the actual stable identifier.

Yes-ish. Any saved metadata that points at the old `snapshot/<id>` branch name will stop resolving once the migration script runs (the underlying commits remain, just under different refs). Anything that records the commit hash (which all callers do — `ExecutionStamp.commit_hash` is propagated alongside the ref) is unaffected.


scripts/migrate_snapshot_branches_to_runs_refs.sh
danbraunai-goodfire, line 1: I'd prefer if this wasn't in the repo, and just shared around with people (or not even bother with that).

danbraunai-goodfire (approved):
nice. Small comment

PR #497, opened 2026-05-15
Decouple component and CI-fn optimizers

ocg-goodfire:
## Description

Replace the single top-level `lr_schedule` + `grad_clip_norm_{components,ci_fns}` fields on `Config` with two `OptimizerConfig` blocks:

```yaml
components_optimizer:
  lr_schedule: { start_val: 5e-5, fn_type: cosine, ... }
  grad_clip_norm: 0.01
ci_fn_optimizer:
  lr_schedule: { start_val: 1e-3, fn_type: constant }
  weight_decay: 0.1
  betas: [0.9, 0.95]
```

`OptimizerConfig` carries `lr_schedule`, `weight_decay`, `betas`, and `grad_clip_norm`. The training loop in `run_param_decomp.optimize` now builds and steps two `AdamW` instances (one per group) and logs per-group LR at `train/schedules/lr/{components,ci_fn}`.

## Motivation and Context

The "components vs CI-fn" split was already a first-class concept in a few places (separate `grad_clip_norm_*` fields, faithfulness warmup that only touches components), but LR / weight_decay / betas were still shared. Splitting them into one cohesive `OptimizerConfig` per group:

- lets us tune the two groups independently (we've wanted this for a while)
- gathers everything that varies per group into one place
- removes the buried `weight_decay=0` magic constant
- replaces three top-level fields with two structured ones

## How Has This Been Tested?

- `make type` → 0 errors
- `make test` → 483 passed, 14 skipped
- Migration validator (`migrate_to_optimizer_configs`) rewrites old top-level `lr_schedule` / `grad_clip_norm_*` keys into the new layout; `tests/test_distributed.py` and `tests/scripts_run/test_grid_search.py` continue to exercise the legacy path

For freshly-authored configs, yes — the top-level `lr_schedule` and `grad_clip_norm_{components,ci_fns}` fields are gone. Existing wandb-saved configs and any handwritten ones in the old shape are auto-migrated by `migrate_to_optimizer_configs`, so loading old runs still works. All in-tree experiment YAMLs have been updated to the new layout.


danbraunai-goodfire (approved):
lgtm. As discussed, this makes it a bit harder to sweep over lr. Best solution is probably to implement a new feature that allows you to tie two hyperparams together and not do a grid over all combinations of them.

PR #501, opened 2026-05-18
Rework sweeps around SweepGenerator; add view_meta to RunMetadata

ocg-goodfire:
## Description

Replaces the YAML-grid-only sweep system with a `(base_config) -> SweepSpec` generator protocol. Adds `view_meta` to `RunMetadata` for researcher-facing labels surfaced to W&B. Moves `wandb_project` / `wandb_run_name` off `PDConfig` and onto `RunMetadata`.

**New shape:**

```python
@dataclass(frozen=True)
class SweepRun:
    name: str
    config: dict[str, Any]
    view_meta: dict[str, Any]  # e.g. {"lr_ratio": 0.1, "size": "medium"}

@dataclass(frozen=True)
class SweepSpec:
    description: str
    runs: list[SweepRun]

class SweepGenerator(ABC):
    name: ClassVar[str]   # for auto-discovery
    def __call__(self, base_config) -> SweepSpec: ...
```

Built-in `CartesianGridSweep` covers the 80% case (flat dot-pathed grid YAML). Custom generators land in `param_decomp/sweeps/` for auto-discovery, or are referenced by `module.path:Class[:arg]`.

**New runner flow:**
1. Resolve `--sweep <spec>` to a generator instance.
2. Call generator → `SweepSpec`.
3. Validate every config against the driver's pydantic type on the launch node (fail fast).
4. Snapshot the materialized `SweepSpec` to `PARAM_DECOMP_OUT_DIR/sweeps/<launch_id>/spec.yaml`.
5. Submit SLURM array.

**view_meta:** new free-form field on `RunMetadata`. Populated by sweep generators (cartesian auto-fills it with the varying axes). Surfaced to W&B under `view_meta/<key>` in `wandb.config` so the UI can group/color by researcher-facing axes — not raw config fields. Empty for one-off / notebook runs.

**`PDConfig` cleanup:** `wandb_project` and `wandb_run_name` removed. They affect observation, not training, and belong on `RunMetadata`. All PD experiment YAMLs updated (these fields removed). `pd-run --project` now writes to metadata. The 4 `config.wandb_project` checks in `optimize()` collapse to `wandb.run is not None`.

## Motivation and Context

The previous sweep system forced a single grid schema with awkward `global:` + per-experiment merge semantics. Conditional axes, paired axes, log-spaces, and "semantic axes" (e.g. one `size` axis controlling several config fields together) all required ugly workarounds. A generator protocol expresses all of these trivially while keeping the cartesian grid case as a one-import built-in.

`view_meta` exists because sweep axes are often conceptual (`lr_ratio`, `size=medium`) rather than direct config-field overrides. Trying to recover those from raw configs in W&B's UI is painful; letting the generator declare them once at submission is cheap and clear.

Moving wandb fields off `PDConfig` is incidental cleanup: `PDConfig` should be hyperparameters, not observation knobs. `RunMetadata` already exists as the right home — it already holds the driver path for the same reason.

Stacked on `feature/modular`.

## How Has This Been Tested?

- [x] `make check` clean (ruff lint, ruff format, basedpyright — 0 errors / 0 warnings)
- [x] `make test` passes (120/120 selected, 7.2s)
- [x] New tests: `tests/sweeps/test_cartesian.py` (7 cases) and `tests/sweeps/test_discovery.py` (4 cases) covering grid expansion, view_meta auto-population, run-name formatting, CLI-string resolution (short name / yaml-path shorthand / unknown-name rejection)
- [x] `tests/scripts_run/test_main.py` rewritten against the new launch flow (sweep array sizing + single-run path)
- [x] Existing `tests/test_tms.py`, `tests/test_resid_mlp.py`, `tests/test_gpt2.py` updated for the removed `PDConfig` fields and still pass

Not yet exercised on a real cluster — SLURM submission paths are mocked. Worth a smoke run before merge.

Yes.

- **`PDConfig.wandb_project` / `PDConfig.wandb_run_name` removed.** Notebook callers that constructed `PDConfig(wandb_project=...)` must set the fields on `RunMetadata` instead and pass it to `run_pd(..., metadata=...)`. All in-tree experiment YAMLs have been migrated; external YAMLs that set `pd.wandb_project` will fail pydantic validation (PDConfig forbids extras).
- **`pd-run --sweep` semantics changed.** No more boolean flag and no more implicit `param_decomp/scripts/sweep_params.yaml` default. Pass a yaml path (`--sweep my_grid.yaml`) or a generator spec (`--sweep cartesian:my_grid.yaml`, `--sweep my_pkg:MyClass`).
- **Sweep YAML schema changed.** Old `{values: [...]}` leaf format and `global:` + per-experiment blocks are gone. New schema is flat dot-pathed:
  ```yaml
  description: "..."
  grid:
    pd.seed: [0, 1, 2]
    pd.loss_metrics.importance_minimality.coeff: [0.1, 0.2]
  ```
- **Saved-run artifact change.** `sweep_params.yaml` is no longer written next to each run's checkpoint. The materialized spec for the whole sweep lives at `PARAM_DECOMP_OUT_DIR/sweeps/<launch_id>/spec.yaml`.
- **`RunMetadata` gains optional fields.** Reading older `run_metadata.yaml` files still works (missing fields default to None / empty dict).


ocg-goodfire:
Folded into #504, which now targets feature/modular with the full reorganization.

PR #503, opened 2026-05-18
Split LoggingConfig out of PDConfig

ocg-goodfire:
## Description

Stacks on #501. Moves 8 fields out of `PDConfig` into a new sibling `LoggingConfig`:

```
train_log_freq, eval_freq, slow_eval_freq, slow_eval_on_first_step,
n_eval_steps, eval_batch_size, save_freq, eval_metrics
```

All eight are observation knobs — they control *how often* / *with what cadence* you peek at a run, not what the run computes. Two runs with identical `PDConfig` and different `LoggingConfig` produce **bit-identical trained models**.

```python
class ExperimentConfig(BaseConfig):
    pd: PDConfig
    logging: LoggingConfig    # new
```

`optimize()` and `run_pd()` take both as separate args. The `loss_metrics` / `eval_metrics` overlap validator moves to `ExperimentConfig` (needs both sides); the `slow_eval_freq % eval_freq == 0` validator moves to `LoggingConfig`. W&B config dump nests logging-side keys under a `logging/` prefix to avoid colliding with PDConfig keys.

## Motivation and Context

Direct follow-up to PR #501. Same lens: keep `PDConfig` as the algorithm config, push everything else out. `wandb_project` / `wandb_run_name` went to `RunMetadata` last PR; the observation cadence + eval-only metrics go to `LoggingConfig` here.

Concrete wins:
- `PDConfig` shrinks from ~25 fields to ~17 algorithm-only fields
- Sweep grids stop having to reason about whether `eval_freq` is "a hyperparameter" (it isn't)
- Reload semantics get cleaner — same algorithm with different observation cadence is obviously the same experiment

## How Has This Been Tested?

- [x] `make check` clean
- [x] Full test suite: 455 passed, 7 skipped
- [x] All 12 in-tree PD experiment YAMLs migrated by script and validated by reload
- [x] Tests that construct `PDConfig` inline (test_tms, test_resid_mlp, test_gpt2, test_component_model, test_server_api, test_run_metadata_round_trip) updated to also construct `LoggingConfig` and pass it to `optimize()`

Not exercised on a real cluster — w&b prefix logging is unverified visually.

Yes.

- **8 fields removed from `PDConfig`.** Any external YAML or `PDConfig(...)` instantiation that sets `train_log_freq` / `eval_freq` / `slow_eval_freq` / `slow_eval_on_first_step` / `n_eval_steps` / `eval_batch_size` / `save_freq` / `eval_metrics` will fail pydantic validation. Move them into a `LoggingConfig` block alongside `pd:` in the experiment YAML, or pass `logging_config=...` to `run_pd` / `optimize`.
- **`run_pd` and `optimize` signatures gain a required `logging_config: LoggingConfig` parameter.** Notebook callers must construct and pass one.
- **`init_wandb` keyword arg `view_meta` and new `extra_configs` are now keyword-only.** Existing callers (run_pd, train_resid_mlp) already pass them as kwargs.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,11 +89,18 @@ README's "Custom experiments" section walks through both routes side-by-side.
 
 ### Per-experiment Configs
 
-Built-in YAML configs are pure experiment configs nested under `pd:`, `target:`, and `data:`:
+Built-in YAML configs are pure experiment configs nested under `pd:`, `logging:`, `target:`,
+and `data:`:
 
-- `LMExperimentConfig(pd, target: LMTargetConfig, data: LMDataConfig)`
-- `TMSExperimentConfig(pd, target, data)`
-- `ResidMLPExperimentConfig(pd, target, data)`
+- `LMExperimentConfig(pd, logging, target: LMTargetConfig, data: LMDataConfig)`
+- `TMSExperimentConfig(pd, logging, target, data)`
+- `ResidMLPExperimentConfig(pd, logging, target, data)`
+
+`PDConfig` carries algorithm hyperparameters (seed, ci_config, losses, optimizers, …) — the
+things that change what the trained model is. `LoggingConfig` carries observation-only
+settings (log/eval/save cadence, eval-only metrics, eval_batch_size) — fields that *don't*
+affect the trained model. Two runs with identical `PDConfig` and different `LoggingConfig`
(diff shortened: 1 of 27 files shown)


ocg-goodfire:
Folded into #504, which now targets feature/modular with the full reorganization.

PR #515, opened 2026-05-23
Add pd-lm-layerwise: per-matrix LM PD launcher

ocg-goodfire:
## Description

New `pd-lm-layerwise` CLI that splits a single LM experiment YAML into one config per `(layer × module_pattern)` pair and submits them as a SLURM array of `pd-lm` jobs.

- For each entry in the base config's `pd.decomposition_targets`, replaces the lone `*` with each requested layer index, emits a config with that single resolved target, and queues a `pd-lm <config>` task.
- `--include "q_proj,k_proj"` filters base patterns by substring; `--layers "0,2"` restricts which layer indices get expanded.
- Asserts exactly one `*` per base pattern; refuses `identity_decomposition_targets` (those would need a separate split strategy).
- Generated configs live at `PARAM_DECOMP_OUT_DIR/layerwise/<run_id>/configs/<pattern>.yaml`. Each `pd-lm` task then writes its own decomposition output under `PARAM_DECOMP_OUT_DIR/decompositions/`.
- Per-task SLURM comments are set to the matrix name for easy `squeue` inspection.
- Uses the existing `create_git_snapshot` / `SlurmArrayConfig` / `submit_slurm_job` infra so it behaves like the other `pd-*` launchers.

## Motivation and Context

MVP for the layerwise scheme we discussed in the team DM — Dan suggested just doing a job-per-matrix to skip the memory pressure of joint training while we iterate. We give up full-model evals in exchange for many independent fast runs.

Targets `feature/modular` because the new module lives under `param_decomp_lab/` and depends on infra that isn't on `main` yet.

## How Has This Been Tested?

- Imported and ran `_build_configs` against `pile_llama_simple_mlp-4L.yaml`:
  - Default expansion → 24 configs (4 layers × 6 patterns), each with a single resolved `decomposition_targets` entry.
  - `--include q_proj,k_proj --layers 0,2` → 4 configs.
- YAML roundtrip: wrote a generated config to disk and reloaded it via `LMExperimentConfig.from_file` — equal to the in-memory config.
- Rendered the SLURM array script (with a fake snapshot ref) and inspected the output: correct `--array=1-N`, per-task `scontrol update` comment block, workspace setup, and one `pd-lm <path>` per case branch.
- `basedpyright param_decomp_lab/experiments/lm/layerwise.py`: 0 errors, 0 warnings.
- `pd-lm-layerwise --help` renders correctly via Fire.

Not yet submitted to a real SLURM array — happy to do that in a follow-up if you'd like a smoke run before merging.

No — purely additive (new module + new console entry + one CLAUDE.md row).


danbraunai-goodfire (approved):
Heh, love how this is just a new script in param_decomp_lab.

PR #516, opened 2026-05-25
Replace SavedRun dispatch with per-experiment SavedXRun classes

ocg-goodfire:
## Description

Replace the polymorphic `SavedRun` + `RunMeta` + `_RUN_MODULE_PATHS` + `experiment_kind` machinery with three concrete frozen dataclasses (`SavedTMSRun` / `SavedResidMLPRun` / `SavedLMRun`) defined alongside each experiment's `run.py`.

The previous design paid a polymorphism bill that no caller was sending: every consumer ended up writing `assert run.kind == "lm"` plus an `isinstance` narrowing on `target_cfg` / `data_cfg`. Concrete classes give callers a fully-typed `cfg: <Name>ExperimentConfig` directly. As a side effect:

- `run_meta.yaml` drops the `experiment_kind` sibling key — it's now just `cfg.to_file(out_dir / RUN_META_FILENAME)`.
- The two-stage validation (eager `pd / runtime / cadence`, raw-dict `target` / `data` validated later) goes away — pydantic validates the whole tree at `from_file` time against the concrete `<Name>ExperimentConfig`.
- Three sources of truth for "what experiment kinds exist" (the `RunKind` literal, the `_RUN_MODULE_PATHS` dict, the per-experiment `save_run_meta(kind="...")` calls) collapse to one (the concrete class per `run.py`).
- A small `load_component_model(pd_config, checkpoint_path, target, run_batch)` helper in `component_model_io.py` single-sources the `PDConfig → ComponentModel construction args` mapping so the three `SavedXRun.load_model()` methods don't duplicate it.

Falls out from the LM-only-postprocessing assumption:
- Clustering simplifies to LM-only: `HarvestConfig` drops `n_samples`, `collect_memberships_resid_mlp` and the dispatch wrapper are deleted, three stale resid_mlp clustering configs are removed, `pretrain_info`'s soft-fail goes away.
- `HarvestConfig.model_path` migrates from `str` + custom `field_validator` to `ModelPath` (now also accepts local checkpoint paths, consistent with the lab's other path fields).
- Stale `wandb:` prefix stripped from in-repo clustering configs — already rejected by `parse_wandb_run_path`, just silently mishandled by `ModelPath` before this fix.

## Related Issue

N/A

## Motivation and Context

This is a follow-up to the `feature/modular` "sexy refactor" PR (#486) addressing the weakest piece of that PR's design: the kind-dispatch machinery in `param_decomp_lab/saved_run.py`. Per discussion: every reload path knows concretely which experiment it's loading, so the polymorphism isn't paying off — it's just shifting kind-checking from compile time to runtime asserts at every call site.

Net: −185 LOC, simpler types, no behavior change for the supported (LM) paths.

## How Has This Been Tested?

- `make check` clean (basedpyright 0 errors, ruff lint+format clean).
- 410/410 tests pass.
- Each of the 4 in-repo LM clustering JSON configs validates end-to-end via `HarvestConfig.from_file`.
- No remaining references to deleted symbols (`SavedRun`, `RunMeta`, `RunKind`, `_RUN_MODULE_PATHS`, `experiment_kind`, `save_run_meta`, `TARGET_CONFIG_TYPE`, `DATA_CONFIG_TYPE`) in either source or tests.

Yes — consistent with the parent PR's no-legacy-shim policy:

- **Saved runs**: `run_meta.yaml` no longer carries `experiment_kind`. Old run metas need a one-key removal or retrain — pydantic will reject the unknown field under `extra="forbid"`.
- **Public API**: post-processing callers import the concrete `Saved<Name>Run` directly from its experiment module (`from param_decomp_lab.experiments.lm.run import SavedLMRun`), not a polymorphic class from `param_decomp_lab.saved_run` (deleted).
- **Clustering `HarvestConfig`**: `n_samples` field removed; `n_tokens` is now required; `wandb:` prefix on `model_path` is no longer accepted (was already broken — `parse_wandb_run_path` rejected it explicitly).


param_decomp_lab/experiments/lm/run.py
```diff
@@ -163,12 +165,77 @@ def make_run_batch(target_cfg: LMTargetConfig) -> RunBatch:
...
+        """Rebuild a `DataLoader` for the requested split."""
+        return build_loader(
+            self.cfg.target,
+            self.cfg.data,
+            split=split,
+            device=device,
+            batch_size=batch_size,
+            dist_state=dist_state,
+            seed=seed,
+        )
```
danbraunai-goodfire: Eh. Might just prefer calling the module-level build_loader directly with the cfg args. Probably should be called build_lm_loader too.

param_decomp_lab/experiments/lm/run.py
```diff
@@ -163,12 +165,77 @@ def make_run_batch(target_cfg: LMTargetConfig) -> RunBatch:
...
+    def cfg_from_path(cls, path: ModelPath) -> LMExperimentConfig:
+        """Load just ``run_meta.yaml`` without resolving the checkpoint.
+
+        Useful for app endpoints that only need config introspection (e.g. the run
+        picker showing architecture summaries) and want to avoid the W&B checkpoint
+        download that ``from_path`` triggers.
+        """
+        return LMExperimentConfig.from_file(
+            resolve_config_path(path, config_filename=RUN_META_FILENAME)
+        )
```
danbraunai-goodfire: This is a bit awkward. Think I'd prefer to just call LMExperimentConfig.from_file at the call sites of make a module-level function that does this rather than a classmethod that returns a different object than the class.

danbraunai-goodfire (approved):
Should we just delete SavedTMSRun and SavedResidMLPRun? They aren't called anywhere.

Apart from that, a few small comments but please merge yourself unless doing anything controversial.

ocg-goodfire:
> Should we just delete SavedTMSRun and SavedResidMLPRun? They aren't called anywhere.

Thought about this. but left them in because I though we'd inevitably want them again. They would've been used if  #486 hadn't deleted a bunch of scripts

param_decomp_lab/experiments/lm/run.py
ocg-goodfire: Done in b0147dd — dropped the method from all three SavedXRun classes and renamed the free functions to `build_{lm,tms,resid_mlp}_loader`. The 3 LM call sites now call the free function directly with `pd_run.cfg.target` / `pd_run.cfg.data`.

param_decomp_lab/experiments/lm/run.py
ocg-goodfire: Done in b0147dd — moved to a module-level `load_lm_cfg(path) -> LMExperimentConfig` and updated the two callers (pretrain_info.py, run_registry.py).

PR #518, opened 2026-05-26
Add 1-pool resumption: Trainer class + canonical training-state checkpoints

ocg-goodfire:
## What this does

Training runs can now be paused and resumed. A run periodically saves enough state to disk that a separate process can pick it up and continue training as if nothing happened — bit-exact on the trainer side.

```bash
pd-lm config.yaml                   # fresh run (also saves resume checkpoints now)
pd-lm --resume my_resume.yaml       # continue from a saved checkpoint
```

Where `my_resume.yaml` is a tiny config pointing at the parent run + which step to resume from.

## Why

Long training jobs get preempted, slurm reservations expire, hyperparameters need tweaking mid-trajectory. The previous training loop had no notion of "where am I" beyond the live `optimize()` call — this lifts that into an explicit, snapshottable `Trainer`.

## The shape of the solution

**One canonical training-state file per checkpoint.** The persistence layer is deliberately topology-free: rank 0 writes a single `training_<step>.pth` containing everything needed to reconstruct the trainer (cfgs, model state, optimizer state, metric states, step). On resume, that one file rebuilds a trainer for whatever topology you're running — no per-rank shards, no layout fingerprint baked into the artifact.

For 1-pool DDP this is straightforward: every rank has the same state replicated, so rank 0's write captures all of it. For 2/3-pool ports later, the trainer's `snapshot()` does the cross-rank gathers needed to produce the same canonical artifact.

**`Trainer` class** (carved out of the old free-function `optimize()`). Holds the step counter, model, optimizer, and metric state. Exposes `snapshot()` to grab a point-in-time view and `from_snapshot(...)` to rebuild a trainer from one. `optimize()` survives as a thin wrapper for callers that don't care about resumption.

**`TrainerSnapshot`** is small: `step` + `state` (the canonical training-state dict). One field for everything, no rank-local/global mixing.

**`RunSink.checkpoint(snapshot)`** — Protocol method now takes a `TrainerSnapshot` instead of `(state_dict, step)`. The lab sink writes two files per checkpoint:

- `model_<step>.pth` — just the component-model state dict. This is what downstream tools (`SavedLMRun.load_model`, postprocessing, the app) already read; the shape is unchanged.
- `training_<step>.pth` — the rest: cfgs, optimizer state, metric states, step. Used only for resumption.

Splitting into two files keeps the "just give me the trained model" path cheap and unchanged for everything else in the codebase.

**Optimizer state is keyed by parameter name.** Rather than persisting PyTorch's integer-indexed optimizer state (where index 5 means "the 6th param the optimizer was constructed with" — topology-dependent), the canonical file stores `{param_name: state_entry}` (e.g. `components.h.0.attn.q_proj.V`, `ci_fn.embed.weight`). On resume, each rank looks up the state for the params it owns by name. Same shape on disk regardless of pool count or world size; resume can pick a different topology and the optimizer state still lands correctly.

**Metric state.** Most metrics are stateless across the optimizer trajectory (accumulators reset between eval passes), but `PersistentPGDReconLoss` carries adversarial sources + Adam moments that have to round-trip. `Metric` now has `state_dict` / `load_state_dict` mirroring the `nn.Module` / `Optimizer` convention; default is a no-op empty dict.

## Out of scope

2-pool and 3-pool resumption. Those subsystems live on a separate branch (`feature/resumption`) and aren't on `feature/modular` yet — they'll be ported onto this template when the time comes. The persistence shape (one canonical state, topology-free, name-keyed optimizer state) is the same; what they add is the cross-rank gather inside `snapshot()` and the scatter inside `from_snapshot()`.

## Breaking

- `RunSink.checkpoint(state_dict, step)` → `RunSink.checkpoint(snapshot: TrainerSnapshot)`. External Protocol implementations need to switch to reading `snapshot.state["component_model"]` (the model state) and `snapshot.step`.
- `optimize()` keeps its signature; existing callers are unchanged.

## Testing

- `test_trainer_snapshot_round_trips`: snapshot + from_snapshot produces matching trainer state including optimizer state.
- `test_trainer_resumes_from_snapshot_and_matches_uninterrupted_run`: resumed run matches an uninterrupted run bit-exact on CPU.
- `test_run_sink_writes_model_and_training_files`, `test_resume_round_trip_matches_uninterrupted_run`, `test_resolve_step_finds_latest`: lab-side integration through the real `RunSink`.
- Full suite: 415 passed, 5 skipped. basedpyright clean.

### DIFF
--- a/param_decomp/metrics/base.py
+++ b/param_decomp/metrics/base.py
@@ -114,3 +114,22 @@ class Metric[TConfig: BaseConfig](ABC):
         Override when a metric needs to step internal state coupled to the outer
         backward — e.g. `PersistentPGDReconLoss` steps its adversarial sources here.
         """
+
+    def state_dict(self) -> dict[str, Any]:
+        """Return persistent metric state for round-tripping across a training restart.
+
+        Default is an empty dict: most metrics are stateless w.r.t. the optimizer
+        trajectory (their accumulators reset between eval passes). Override when a
+        metric carries trajectory-dependent state — e.g. `PersistentPGDReconLoss`
+        round-trips its adversarial sources here. Mirrors the `nn.Module` / `Optimizer`
+        convention; the parent `Trainer` composes these into its own state blob.
+        """
+        return {}
+
+    def load_state_dict(self, state: dict[str, Any]) -> None:  # noqa: B027 — intentional no-op default
+        """Load persistent state produced by a prior :meth:`state_dict` call.
+
+        Default is a no-op (matches the default empty `state_dict`). Override
+        alongside `state_dict` for any metric carrying trajectory-dependent state.
+        """
+        del state
--- a/param_decomp/metrics/persistent_pgd_recon.py
+++ b/param_decomp/metrics/persistent_pgd_recon.py
@@ -7,7 +7,7 @@ state + optimizer state machine live in `persistent_pgd_state`.
 """
 
 from collections.abc import Iterable
-from typing import Annotated, ClassVar, Literal, override
+from typing import Annotated, Any, ClassVar, Literal, override
 
 import torch
 from jaxtyping import Float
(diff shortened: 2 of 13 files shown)


ocg-goodfire:
Reopening with renamed branch (feature/resumption-2 → feature/resumption).

PR #520, opened 2026-05-26
Add 1-pool resumption: Trainer class + canonical training-state checkpoints

ocg-goodfire:
## What this does

Training runs can now be paused and resumed. A run periodically saves enough state to disk that a separate process can pick it up and continue training as if nothing happened — bit-exact on the trainer side.

```bash
pd-lm config.yaml                   # fresh run (also saves resume checkpoints now)
pd-lm --resume my_resume.yaml       # continue from a saved checkpoint
```

Where `my_resume.yaml` is a tiny config pointing at the parent run + which step to resume from.

## Why

Long training jobs get preempted, slurm reservations expire, hyperparameters need tweaking mid-trajectory. The previous training loop had no notion of "where am I" beyond the live `optimize()` call — this lifts that into an explicit, snapshottable `Trainer`.

## The shape of the solution

**One canonical training-state file per checkpoint.** The persistence layer is deliberately topology-free: rank 0 writes a single `training_<step>.pth` containing everything needed to reconstruct the trainer (cfgs, model state, optimizer state, metric states, step). On resume, that one file rebuilds a trainer for whatever topology you're running — no per-rank shards, no layout fingerprint baked into the artifact.

For 1-pool DDP this is straightforward: every rank has the same state replicated, so rank 0's write captures all of it. For 2/3-pool ports later, the trainer's `snapshot()` does the cross-rank gathers needed to produce the same canonical artifact.

**`Trainer` class** (carved out of the old free-function `optimize()`). Holds the step counter, model, optimizer, and metric state. Exposes `snapshot()` to grab a point-in-time view and `from_snapshot(...)` to rebuild a trainer from one. `optimize()` survives as a thin wrapper for callers that don't care about resumption.

**`TrainerSnapshot`** is small: `step` + `state` (the canonical training-state dict). One field for everything, no rank-local/global mixing.

**`RunSink.checkpoint(snapshot)`** — Protocol method now takes a `TrainerSnapshot` instead of `(state_dict, step)`. The lab sink writes two files per checkpoint:

- `model_<step>.pth` — just the component-model state dict. This is what downstream tools (`SavedLMRun.load_model`, postprocessing, the app) already read; the shape is unchanged.
- `training_<step>.pth` — the rest: cfgs, optimizer state, metric states, step. Used only for resumption.

Splitting into two files keeps the "just give me the trained model" path cheap and unchanged for everything else in the codebase.

**Optimizer state is keyed by parameter name.** Rather than persisting PyTorch's integer-indexed optimizer state (where index 5 means "the 6th param the optimizer was constructed with" — topology-dependent), the canonical file stores `{param_name: state_entry}` (e.g. `components.h.0.attn.q_proj.V`, `ci_fn.embed.weight`). On resume, each rank looks up the state for the params it owns by name. Same shape on disk regardless of pool count or world size; resume can pick a different topology and the optimizer state still lands correctly.

**Metric state.** Most metrics are stateless across the optimizer trajectory (accumulators reset between eval passes), but `PersistentPGDReconLoss` carries adversarial sources + Adam moments that have to round-trip. `Metric` now has `state_dict` / `load_state_dict` mirroring the `nn.Module` / `Optimizer` convention; default is a no-op empty dict.

## Out of scope

2-pool and 3-pool resumption. Those subsystems live on a separate branch (`feature/resumption`) and aren't on `feature/modular` yet — they'll be ported onto this template when the time comes. The persistence shape (one canonical state, topology-free, name-keyed optimizer state) is the same; what they add is the cross-rank gather inside `snapshot()` and the scatter inside `from_snapshot()`.

## Breaking

- `RunSink.checkpoint(state_dict, step)` → `RunSink.checkpoint(snapshot: TrainerSnapshot)`. External Protocol implementations need to switch to reading `snapshot.state["component_model"]` (the model state) and `snapshot.step`.
- `optimize()` keeps its signature; existing callers are unchanged.

## Testing

- `test_trainer_snapshot_round_trips`: snapshot + from_snapshot produces matching trainer state including optimizer state.
- `test_trainer_resumes_from_snapshot_and_matches_uninterrupted_run`: resumed run matches an uninterrupted run bit-exact on CPU.
- `test_run_sink_writes_model_and_training_files`, `test_resume_round_trip_matches_uninterrupted_run`, `test_resolve_step_finds_latest`: lab-side integration through the real `RunSink`.
- Full suite: 415 passed, 5 skipped. basedpyright clean.

### DIFF
--- a/param_decomp/metrics/base.py
+++ b/param_decomp/metrics/base.py
@@ -114,3 +114,22 @@ class Metric[TConfig: BaseConfig](ABC):
         Override when a metric needs to step internal state coupled to the outer
         backward — e.g. `PersistentPGDReconLoss` steps its adversarial sources here.
         """
+
+    def state_dict(self) -> dict[str, Any]:
+        """Return persistent metric state for round-tripping across a training restart.
+
+        Default is an empty dict: most metrics are stateless w.r.t. the optimizer
+        trajectory (their accumulators reset between eval passes). Override when a
+        metric carries trajectory-dependent state — e.g. `PersistentPGDReconLoss`
+        round-trips its adversarial sources here. Mirrors the `nn.Module` / `Optimizer`
+        convention; the parent `Trainer` composes these into its own state blob.
+        """
+        return {}
+
+    def load_state_dict(self, state: dict[str, Any]) -> None:  # noqa: B027 — intentional no-op default
+        """Load persistent state produced by a prior :meth:`state_dict` call.
+
+        Default is a no-op (matches the default empty `state_dict`). Override
+        alongside `state_dict` for any metric carrying trajectory-dependent state.
+        """
+        del state
--- a/param_decomp/optimize.py
+++ b/param_decomp/optimize.py
@@ -1,13 +1,25 @@
-"""PD optimization loop.
+"""The PD trainer.
 
-`optimize()` is the sole core entrypoint; `EvalLoop` bundles the eval runtime objects
-with their timing.
+Two entry points:
+
+- :class:`Trainer` — stateful training class. Construction sets up the
+  `ComponentModel`, the two optimizers, and the loss-metric instances from
+  ``pd_config``. :meth:`Trainer.run` advances the training loop from
+  ``self.step`` to ``pd_config.steps``. :meth:`Trainer.snapshot` and
+  :meth:`Trainer.from_snapshot` round-trip an atomic
+  :class:`~param_decomp.trainer_snapshot.TrainerSnapshot` that lets a caller
+  persist and restore the full training state (used by resumption).
+- :func:`optimize` — thin convenience wrapper that constructs a `Trainer` and
+  calls :meth:`Trainer.run`. Preserves the original function-style entry point
+  for callers that don't need resumption.
+
+Both go through the same code path.
 """
 
 import gc
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, Self, cast
 
 import torch
 import torch.nn as nn
@@ -47,19 +59,28 @@ from param_decomp.metrics.persistent_pgd_recon import validate_pgd_scope
 from param_decomp.run_sink import RunSink
 from param_decomp.schedule import get_scheduled_value
 from param_decomp.torch_helpers import bf16_autocast, loop_dataloader
+from param_decomp.training_state import TrainingState
 
 
 @dataclass(frozen=True)
 class EvalLoop:
     """Eval-loop runtime objects bundled with their timing.
 
-    Pass `eval_loop=None` to `optimize` to skip eval entirely. When set, the trainer
-    evaluates every `every` steps; on steps that are also multiples of `slow_every`,
-    slow metrics fire too. `slow_every` must be a multiple of `every` — `should_run_slow_eval`
-    is only checked on steps where `should_eval` already fired.
+    Pass ``eval_loop=None`` to :meth:`Trainer.run` (or :func:`optimize`) to skip
+    eval entirely. When set, the trainer evaluates every ``every`` steps; on steps
+    that are also multiples of ``slow_every``, slow metrics fire too. ``slow_every``
+    must be a multiple of ``every`` — the trainer only checks :meth:`should_run_slow_eval`
+    on steps where :meth:`should_eval` already fired.
 
-    `optimize` calls `Metric.bind(model, device)` on every entry of `metrics` before the
-    loop starts.
+    Attributes:
+        loader: Eval data loader. Looped for the lifetime of training.
+        metrics: Caller-instantiated eval ``Metric``s. ``optimize`` calls
+            ``Metric.bind(model, device)`` on each before the loop.
+        n_steps: Number of eval batches per eval pass.
+        every: Period (in train steps) between eval passes.
+        slow_every: Period (in train steps) between *slow* eval passes. Must
+            be a multiple of ``every``.
+        slow_on_first_step: Whether slow eval fires at step 0.
     """
 
     loader: DataLoader[Any]
@@ -75,13 +96,14 @@ class EvalLoop:
         )
 
     def should_eval(self, step: int) -> bool:
+        """Whether a (regular) eval pass should fire at ``step``."""
         return step % self.every == 0
 
     def should_run_slow_eval(self, step: int) -> bool:
-        """Whether slow eval should fire at `step`.
+        """Whether slow eval metrics should fire at ``step``.
 
-        Slow eval is gated on top of `should_eval`; only call this on steps where
-        `should_eval` is already true.
+        Slow eval is gated on top of ``should_eval``; callers are expected to
+        only call this on steps where ``should_eval`` is already true.
         """
         if step == 0:
             return self.slow_on_first_step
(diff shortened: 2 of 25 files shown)


param_decomp/optimize.py
```diff
@@ -139,250 +184,478 @@ def tie_component_weights(
...
-                clip_grad_norm_(component_params, pd_config.components_optimizer.grad_clip_norm)
-            if pd_config.ci_fn_optimizer.grad_clip_norm is not None:
-                clip_grad_norm_(ci_fn_params, pd_config.ci_fn_optimizer.grad_clip_norm)
-            components_optimizer.step()
-            ci_fn_optimizer.step()
-
-    if is_main_process():
-        logger.info("Finished training loop.")
+    Thin wrapper over :class:`Trainer` for callers that don't need resumption
+    or interactive control. Identical signature to the pre-Trainer ``optimize``.
```
danbraunai-goodfire: Remove this type of comment that talks about code prior to this

param_decomp/optimize.py
```diff
@@ -139,250 +184,478 @@ def tie_component_weights(
...
+                    f"duplicate eval metric {metric_name!r}"
+                )
+                eval_only_instances[metric_name] = m
+            overlap = sorted(set(self.loss_metrics) & set(eval_only_instances))
+            assert not overlap, (
+                f"eval_loop.metrics overlap with pd_config.loss_metrics: {overlap}. Loss "
+                "metrics are automatically evaluated; remove the duplicates from "
+                "eval_loop.metrics."
+            )
+        all_instances = {**self.loss_metrics, **eval_only_instances}
```
danbraunai-goodfire: Can put this in a helper function or instance method

param_decomp_lab/experiments/lm/run.py
```diff
@@ -254,17 +290,91 @@ def main(
...
             reconstruction_loss=recon_loss_kl,
             pd_config=cfg.pd,
             runtime_config=cfg.runtime,
-            sink=sink,
-            cadence=cfg.cadence,
-            eval_loop=eval_loop,
         )
+        trainer.run(train_loader, sink, cfg.cadence, eval_loop)
+    finally:
+        sink.finish()
```
danbraunai-goodfire, line 301: Now that more code happens here, it's tempting to make sing a context manager. But I think I'll push against this. Researchers don't want to have to use context managers for the main entrypoint. Even this try-finally is pretty annoying.

param_decomp_lab/experiments/lm/run.py
danbraunai-goodfire, line 1: I don't like that the resume functionality is specific to the lm experiment. I think the majority of it should be generic. 

I suppose I'm OK with this for now, until we have another concrete experiment type that we care about (e.g. a bio model), at which point we can worry about consolidating this stuff.

I think the submit_slurm functionality has an even bigger case for being generic, it was an oversight by me merging this in. I made #522 for this

param_decomp/metrics/base.py
```diff
@@ -114,3 +114,22 @@ def after_backward(self) -> None:  # noqa: B027 — intentional no-op default
...
+        """
+        return {}
+
+    def load_state_dict(self, state: dict[str, Any]) -> None:  # noqa: B027 — intentional no-op default
+        """Load persistent state produced by a prior :meth:`state_dict` call.
+
+        Default is a no-op (matches the default empty `state_dict`). Override
+        alongside `state_dict` for any metric carrying trajectory-dependent state.
+        """
+        del state
```
danbraunai-goodfire, line 135: I don't like that we now have 4 methods on this base class that are just for PersistentPGDReconLoss, and are likely to only ever be for that loss. Though I think PPGD behaves enough like a "metric" that it probably does make sense to pay some cost here. I don't have a better idea.

param_decomp_lab/resumption/config.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import Any, Literal
+
+import torch
+from pydantic import PositiveInt
+
+from param_decomp.base_config import BaseConfig
+from param_decomp.training_state import TrainingState
+
+
+class ResumeOverrides(BaseConfig):
```
danbraunai-goodfire: I don't think we need any of this overrides stuff in this PR. The standard resume case is that the run gets killed and you just want to continue it with the original config. No override needed.

The less common but plausible case is that someone wants to run more steps on an already completed run. If that happens, they can just do:
```
  snapshot = read_training_snapshot(parent_dir, step)
  snapshot.pd_config["steps"] = 200
  trainer = Trainer.from_snapshot(snapshot, target_model=..., ...)
```

param_decomp/optimize.py
danbraunai-goodfire: I'm unsure whether this wrapper is worth it. I'm tempted to just say there's one entrypoint which involves initializing a Trainer and calling run. Not sure.

danbraunai-goodfire (commented):
There's a lot of code here. On its face most of it makes sense. But it does feel like there might be a cleaner way. Absent any ideas though, happy to go ahead with this.

Please either incorporate here or make an issue that checkpoints when a SIGTERM is received.

I'd also like a feature which deletes old checkpoints when a new one is saved. And I think this should be on by default. Can make in this PR if you agree.

ocg-goodfire:
_(Claude here, replying via Oli's auth after going over the review with him.)_

Thanks for the careful review. Took the three top-level asks seriously:

**On "feels like there might be a cleaner way":** We did a deliberate pass looking for a structural simplification, and our conclusion is that the shape is right — each chunk is paying for something specific (the live `Trainer` / persistable `TrainingState` split for snapshot atomicity; name-keyed optimizer state to survive a topology change on resume; `Metric.state_dict` / `load_state_dict` so PPGD's adversarial sources round-trip; lab-side automation behind the `--resume` entrypoint).

The one structural simplification we considered seriously was collapsing `ResumeConfig` (now just `from_run + step` after the override removal) into CLI flags like `pd-lm --resume <dir> [--resume-step N]` — about ~50 LOC down. We decided against it: Oli has a strong preference for type-checked YAML over CLI flags, and on reflection I agree — the YAML is small but it's the canonical, type-checked declaration of "which run to resume," and it composes naturally with the existing `pd-*` YAML-driven infra rather than splintering into ad-hoc flags.

The volume in this PR is paying for the **lab-side automation behind `--resume`** (loading the parent's `run_meta.yaml`, rebuilding target / loader / eval / sink, writing provenance, distributed setup, step resolution), not for the schema itself. Compared to the manual snippet from your `ResumeOverrides` comment, the user otherwise has to do all 9 of those steps themselves per resumption.

**On SIGTERM-triggered checkpoint:** Done in this PR ([`f153f445`](https://github.com/goodfire-ai/param-decomp/pull/520/commits/f153f445)). Minimal handler that flips a flag; the existing end-of-step save block checks the flag and saves unconditionally (on top of the cadence check), then breaks out of the loop. SLURM sends SIGTERM to all ranks on preempt, so a preempted run now saves cleanly and resumes from the latest step.

**On auto-delete old checkpoints (default on):** Agree this is worth having and it's orthogonal to the rest of this PR. Happy for you to take it in a follow-up as you offered, or we can add it here — your call.

**Inline comments addressed in [`f153f445`](https://github.com/goodfire-ai/param-decomp/pull/520/commits/f153f445):**
- `optimize.py:652` — dropped the "Identical signature to the pre-Trainer optimize" phrase.
- `optimize.py:496` — extracted `Trainer._build_all_metric_instances` helper.
- `resumption/config.py:18` — `ResumeOverrides` removed entirely. Tests updated to use `snapshot.pd_config["steps"] = ...` for the extend-steps edge case (also caught a stale test reference in [`a6af1a96`](https://github.com/goodfire-ai/param-decomp/pull/520/commits/a6af1a96)).

**Read-and-acknowledge, no code change:**
- `run.py:302` try/finally — kept as is, agree the CM tradeoff isn't worth it for researchers.
- `run.py:1` lm-specific resume code — agree on the direction; deferring with #522 until another concrete experiment kind needs it.
- `metrics/base.py:135` — agree it's not ideal that the base has methods for one consumer, but the alternatives (StatefulMetric mixin, PPGD as non-Metric thing on the trainer) all read as worse.

Happy to discuss any of these further if you disagree with the shape we landed on.

param_decomp/optimize.py
ocg-goodfire: yep, done

param_decomp/optimize.py
ocg-goodfire: > Remove this type of comment that talks about code prior to this

Thanks for catching. god i hate that haha

ocg-goodfire: > I'm tempted to just say there's one entrypoint which involves initializing a Trainer and calling run. Not sure.

I think I agree. I'll do this here and we can see how it looks

ocg-goodfire:
(real Oli here)

I think I disagree about deleting old checkpoints. or at least I think there should be an option to not do this. feels like a trivial storage cost which be annoying. I can imagine wanting to fork a run from an earlier checkpoint. especially a long run

danbraunai-goodfire:
> I think I disagree about deleting old checkpoints. or at least I think there should be an option to not do this. feels like a trivial storage cost which be annoying. I can imagine wanting to fork a run from an earlier checkpoint. especially a long run

Yeah happy for there to be an option not to do this. Fwiw I'm probably more averse than I should be to wasting storage.

PR #529, opened 2026-05-27
Compute weight_deltas outside bf16_autocast

danbraunai-goodfire:
## Description

`_build_metric_context` was calling `component_model.calc_weight_deltas()` inside the train and eval `bf16_autocast` blocks. Move it out — both in the function (now accepts an optional pre-computed `weight_deltas` kwarg) and at both call sites in `Trainer`.

## Motivation and Context

Inside `bf16_autocast`, `components.weight = einsum(V, U)` returns bf16, so `delta = target_weight - V@U` carries **bf16 precision** — a residual floor of ~5e-9 for weight matrices of magnitude ~0.01, vs fp32's ~1e-9. `FaithfulnessLoss` has `coeff=1e7` (calibrated against the fp32 floor), so a ~5× larger residual amplifies into ~2× larger component gradients across all training steps. Downstream this shows up as worse ImpMin, higher L0, larger `kl_random_masked`, and visibly spikier gradient norms compared to Jose's `s-55ea3f9b` baseline.

Pre-refactor (commit `74146b55`, just before `param_decomp_lab/` was carved out) computed `weight_deltas` outside autocast; this PR restores that behaviour.

## How Has This Been Tested?

Reduced-scale harness on `pile_llama_simple_mlp-4L` target (2-module decomposition, smaller CI fn, 5k steps) at `74146b55` and at `769afc1e` (post-refactor with the bug) confirmed the regression: post-refactor faithfulness loss sits at ~5-9e-9 throughout training vs pre-refactor's ~1e-9, and post-refactor `train/grad_norms/summary/total` is ~2× pre-refactor's.

Verified the cause is not seed ordering: a control run with `seed_all_ranks(seed)` removed before `ComponentModel.__init__` shows identical metrics to the unpatched post-refactor.

All harness and Jose-reproduction runs are in [this wandb workspace view](https://wandb.ai/goodfire/param-decomp/workspace?nw=imw4tvl19c).

A full Jose-config (`pile_llama_simple_mlp-4L`, 400k steps) run with this fix needs to confirm the trajectory lands on Jose's curves for ImpMin, total_l0, kl_random_masked, and grad_norm/total at parity step counts.

No. The optional `weight_deltas` kwarg on `_build_metric_context` defaults to recomputing in-place, so direct callers of that internal helper see no behaviour change.

### DIFF
--- a/param_decomp/optimize.py
+++ b/param_decomp/optimize.py
@@ -140,7 +140,7 @@ def _build_metric_context(
     component_model: ComponentModel,
     config: PDConfig,
     reconstruction_loss: ReconstructionLoss,
-    weight_deltas: dict[str, Tensor] | None = None,
+    weight_deltas: dict[str, Tensor],
 ) -> MetricContext:
     # The wrapped_model(...) call here is what registers DDP gradient hooks for this step.
     # Required even if no metric uses the DDP wrapper directly.
@@ -151,8 +151,6 @@ def _build_metric_context(
         detach_inputs=False,
         sampling=config.sampling,
     )
-    if weight_deltas is None:
-        weight_deltas = component_model.calc_weight_deltas()
     return MetricContext(
         model=component_model,
         batch=batch,
--- a/param_decomp/optimize.py
+++ b/param_decomp/optimize.py
@@ -151,7 +151,6 @@ def _build_metric_context(
         detach_inputs=False,
         sampling=config.sampling,
     )
-    # NOTE: FaithfulnessLoss is sensitive to the precision of the weight deltas.
     if weight_deltas is None:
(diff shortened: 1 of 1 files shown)


param_decomp/optimize.py
```diff
@@ -150,7 +151,8 @@ def _build_metric_context(
         detach_inputs=False,
         sampling=config.sampling,
     )
-    weight_deltas = component_model.calc_weight_deltas()
+    if weight_deltas is None:
+        weight_deltas = component_model.calc_weight_deltas()
```
ocg-goodfire: feels like this should be required and the failover removed?

danbraunai-goodfire: Oops yep ty

PR #545, opened 2026-06-02
fix(3-pool): scale PPGD's CI grad by n_ci to survive the CI-pool AVG

danbraunai-goodfire:
## Problem

The CI pool finishes each step with an **AVG all-reduce over its `n_ci` ranks** (`all_reduce_ci_fn_grads`, `step_ci.py`). Every per-position CI gradient is injected on `lower_leaky` of the single CI rank that owns that batch position, so the AVG divides each contribution by `n_ci`. The three sources feeding the CI fn must each compensate:

| Source into CI fn | `n_ci` compensation |
|---|---|
| imp-min | ✅ via the autograd-aware `dist.nn.functional.all_reduce` backward (it SUM-reduces the grad → ×`n_ci`) |
| LW stoch | ✅ explicit, `denom = … * n_per_block / n_ci` |
| **PPGD** | ❌ reused the V/U scale `coeff_ppgd / (n_examples_local * n_ppgd)` — **no `n_ci`** |

So PPGD's gradient *into the CI function* was `1/n_ci` of intended. **V/U is unaffected** (it's SUM-reduced in the PPGD pool then in-block-AVG'd on LW — it never hits the CI-pool AVG), so only the adversary's influence on the CI fn was wrong.

Impact: no-op at `n_ci=1` (the 8-GPU example + current tests, which is why it was invisible), but ~**16–24× too weak** at production CI-pool sizes (`n_ci` = 16 for b256, 24 for gpt2-xl). This silently shifts the stoch:imp:ppgd balance in the CI update.

## Fix

`step_ppgd._scale_grads`: split the V/U and CI scales, multiplying the CI grad by `n_ci` (mirroring the LW stoch path). One line of behavior change + docstring/comment updates.

## Tests

- Confirmed the mechanism with a 2-rank gloo reproduction: a PPGD-style grad injected per-slice + CI-pool AVG comes out at `ref/n_ci`; with the `×n_ci` factor (and the imp-min `dist_fn` path) it matches the single-process reference exactly.
- Added `test_three_pool_grad_scaling.py`: asserts the CI grad now carries the extra `×n_ci` over V/U, sources stay `1/n_examples_local`, and CI==V/U collapses at `n_ci=1`. Fails on the pre-fix code.

### DIFF
--- a/param_decomp_lab/tests/test_three_pool_grad_scaling.py
+++ b/param_decomp_lab/tests/test_three_pool_grad_scaling.py
@@ -1,0 +1,52 @@
+"""Regression test for PPGD -> CI gradient scaling under a multi-rank CI pool.
+
+The CI pool ends each step with an AVG all-reduce over its ``n_ci`` ranks
+(``all_reduce_ci_fn_grads``). PPGD's CI grad is injected per-position on a single
+CI rank, so that AVG divides it by ``n_ci``; ``_scale_grads`` must pre-multiply
+the CI grad by ``n_ci`` to compensate (the LW stoch path does the same via its
+``/ n_ci`` denom). V/U never hits that AVG, so it keeps the plain scale.
+
+At ``n_ci=1`` the factor is a no-op — which is why the bug was invisible to the
+8-GPU (``n_ci=1``) configs and only bit production (``n_ci`` = 16 / 24).
+"""
+
+from types import SimpleNamespace
+from typing import cast
+
+import torch
+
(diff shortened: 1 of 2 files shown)


ocg-goodfire:
@claude can you please explain why we even need to manually scale grads here?

ocg-goodfire:
@claude is there not a simpler fix available? instead of taking a reduction and accounting for that, just ship raw gradients to the correct place?

PR #551, opened 2026-06-05
feat: prune old checkpoints from wandb alongside local disk

danbraunai-goodfire:
## Description

`keep_last_n_checkpoints` now prunes the wandb run as well as local disk. Previously it only `unlink`ed local files, so every checkpointed step persisted on wandb forever.

The prune is split into:
- `_checkpoint_steps_to_prune` — pure step selector (the only real logic; unit-tested).
- `_delete_wandb_files` — best-effort remote deletion via the wandb public API, using `run.files(names=...)` for server-side filtering (existing matches only; no full file-list scan).
- `_prune_old_checkpoints` — thin orchestrator: local unlink + optional remote delete.

## Motivation and Context

A run with `keep_last_n_checkpoints: 2` kept only the last two `(model, training)` pairs on local disk but left all of them on wandb (e.g. `p-ec62b8cb` accumulating `model_25000/50000/75000.pth`). Disk footprint was bounded; wandb storage was not.

**Behavior change worth flagging:** wandb is no longer a durable backup of pruned checkpoints. With `keep_last_n_checkpoints` set, the remote now also keeps only the last N. Runs without that field set are unaffected (pruning is opt-in via the field, as before).

The remote delete uses a deliberately broad `except Exception`: the wandb public API raises errors that `normalize_exceptions` does **not** coerce to `CommError` (unwrapped `run.files()` pagination, non-`CommError` `Error` subclasses), so a narrow catch would let a flaky delete crash a multi-day training run. Pruning is best-effort cleanup, so failures are logged and swallowed; `BaseException` (Ctrl-C, `SystemExit`) still propagates.

## How Has This Been Tested?

- New unit test `test_checkpoint_steps_to_prune_selects_oldest_beyond_keep_last_n` covers the step selector (oldest-first, model/training union, junk-name skipping, keep-all cases).
- Existing `test_keep_last_n_checkpoints_prunes_older_pairs` still passes (local pruning unchanged).
- basedpyright + ruff clean.
- The remote-deletion error handling was validated by simulating the realistic failure modes (`ConnectionError`/timeout from paginator iteration, non-`CommError` `Error` subclasses) and confirming a narrow `CommError`-only catch lets them escape while the broad catch contains them. The remote path itself can't be unit-tested without mocking the public API.

Not API-breaking, but a behavior change: wandb checkpoints are now pruned to the same `keep_last_n_checkpoints` window as local disk, rather than being retained indefinitely. This is irreversible per-file. Only affects runs that set `keep_last_n_checkpoints`.

### DIFF
--- a/param_decomp_lab/tests/test_resumption.py
+++ b/param_decomp_lab/tests/test_resumption.py
@@ -29,7 +29,7 @@ from param_decomp_lab.resumption import (
     read_training_snapshot,
     resolve_step,
 )
-from param_decomp_lab.run_sink import RunSink
+from param_decomp_lab.run_sink import RunSink, _checkpoint_steps_to_prune
 
 
 class TinyLinear(nn.Module):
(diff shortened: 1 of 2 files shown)


param_decomp_lab/tests/test_resumption.py
ocg-goodfire, line 1: @claude is there a nice holistic refactor available here that stops this being a private function test?

ocg-goodfire (approved):

param_decomp_lab/tests/test_resumption.py
ocg-goodfire: @claude are you there?

ocg-goodfire:
@claude  is there a nice holistic refactor available here that stops this being a private function test?

PR #554, opened 2026-06-09
feat(metrics): log beta-independent importance-minimality (no_beta) term

danbraunai-goodfire:
## Summary
`ImportanceMinimalityLoss.compute()` now returns two values: the existing headline term and a new `{name}/no_beta` term — the pure L_p term evaluated at `beta=0`.

`no_beta` is a beta-independent sparsity proxy. Unlike the headline importance-minimality loss (which folds in the configured `beta`) and unlike L0, it stays comparable across experiments that tune `beta`, making it a stable sparsity yardstick.

## Test plan
- `make check` (ruff + basedpyright) passes.
- `python -m pytest param_decomp/tests/metrics/test_importance_minimality_loss.py` — 12 passed (includes new coverage for the `no_beta` key).

### DIFF
--- a/param_decomp/metrics/importance_minimality.py
+++ b/param_decomp/metrics/importance_minimality.py
@@ -67,20 +67,33 @@ def _per_component_sums(
     return out, n_examples
 
 
+def _lp_and_entropy_terms(
+    per_component_sums: dict[str, Float[Tensor, " C"]],
+    n_examples: int,
+    world_size: int,
+) -> tuple[Float[Tensor, ""], Float[Tensor, ""]]:
+    """The two additive parts of the loss, summed over components: `(lp, entropy)`.
+
+    Full loss is `lp + beta * entropy`; `lp` alone is the beta-independent sparsity proxy.
+    """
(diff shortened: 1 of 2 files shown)


param_decomp/metrics/importance_minimality.py
```diff
@@ -164,9 +177,9 @@ def compute(self) -> MetricResult:
...
-            n_examples=n_examples,
-            beta=self.cfg.beta,
-            world_size=1,
-        )
+        lp, entropy = _lp_and_entropy_terms(reduced_sums, n_examples, world_size=1)
+        name = type(self).__name__
+        return {
+            name: lp + self.cfg.beta * entropy,
+            f"{name}_no_beta": lp,
+        }
```
ocg-goodfire, line 185: @claude I think I'm missing something. why is this change from a tensor to a dict valid?

ocg-goodfire:
@claude  I think I'm missing something. why is the change from a tensor to a dict return valid here?

ocg-goodfire (approved):
very nice

PR #556, opened 2026-06-09
perf(multipool): fuse PPGD source-grad into the main backward

danbraunai-goodfire:
## Summary
The final PPGD forward was backproped **twice** per step: `before_backward` ran a separate `autograd.grad(recon_loss, sources, retain_graph=True)` (a full backward through the recon forward) and then `total_loss.backward()` ran another full backward for the component grads. The sources are leaves in the `total_loss` graph, so the main backward already populates `source.grad` — the separate pass is redundant.

This drops it: `before_backward` just flags the step, and `after_backward` harvests `source.grad / coeff` via a new `PersistentPGDState.collect_source_grads`, then runs the **unchanged** `reduce_source_grads` (replica-sync AVG) + `step`. `total_loss` adds the recon term scaled by `coeff` and nothing else depends on the sources, so `source.grad == coeff · ∂recon/∂source`; dividing by `coeff` recovers exactly what `get_grads(recon_loss)` returned.

Saves ~one full backward per step. This is the multipool port of the same change measured at **~7% faster** on the base 4L path (dp2, batch 16).

## Equivalence
Gradients are identical, not approximate — only *how the raw per-rank grad is obtained* changes; `reduce_source_grads` and `step` are untouched. New unit test `test_collect_source_grads_matches_get_grads` asserts `source.grad/coeff` == the separate `get_grads` pass.

## Test plan
- `make check` clean; full PPGD suite (`test_spd_losses.py`) passes (35, incl. the new test).
- ⚠️ **Not run locally:** the 8-GPU distributed pooled grad-check (`test_three_pool_grad_check_distributed`, `@slow`, `nproc_per_node=8`) — this node has 1 GPU. Submitting it on an 8-GPU node to validate the replica-sync path; will post the result here. The replica-sync logic is unchanged, so it should pass given the unit-test equivalence.

### DIFF
--- a/param_decomp/metrics/persistent_pgd_recon.py
+++ b/param_decomp/metrics/persistent_pgd_recon.py
@@ -30,7 +30,6 @@ from param_decomp.metrics.persistent_pgd_state import (
     PersistentPGDSourceScope,
     PersistentPGDState,
     PGDOptimizerConfig,
-    PPGDSources,
     RepeatAcrossBatchScope,
     get_ppgd_mask_infos,
     scope_needs_replica_sync,
@@ -129,7 +128,7 @@ class _PersistentPGDReconBase[
     def __init__(self, cfg: TConfig) -> None:
         super().__init__(cfg)
         self.state: PersistentPGDState | None = None
-        self._pending_source_grads: PPGDSources | None = None
+        self._source_step_active = False
         # Stash from `load_state_dict` if called before the first `update()` —
         # `PersistentPGDState` needs batch_dims, which we only learn from a live ctx.
         self._pending_resume_state: dict[str, Any] | None = None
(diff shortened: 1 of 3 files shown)


danbraunai-goodfire:
✅ **8-GPU distributed pooled grad-check passed** (`test_three_pool_grad_check_distributed`, `--runslow`, `nproc_per_node=8`, job 649369): `1 passed in 87.29s`, SLURM `COMPLETED 0:0`.

Confirms the fully-reduced CI-fn and V/U gradients at the non-square topology (all loss terms incl. PPGD recon) remain bit-identical to the single-process full-batch reference with the fused backward — i.e. dropping the separate source backward doesn't perturb the component-grad path. Combined with the unit-test equivalence of the source grads themselves (`source.grad/coeff == get_grads`) and the unchanged `reduce_source_grads`/`step`, the change is gradient-equivalent end-to-end.

PR #558, opened 2026-06-11
feat(metrics): allow one metric class as both a loss and an eval probe

danbraunai-goodfire:
## Problem

Metric instances are keyed by class name everywhere — instance dicts, state-dict, and log-key suffixes. So the same metric class can't appear under both `pd.loss_metrics` and `eval.metrics`: the instantiation overlap check rejects it. This blocks a natural setup — a cheap 1-step PGD training loss plus a stronger 20-step PGD eval probe — since both are `PGDReconLoss`.

## Fix

- Add optional `name` to `LossMetricConfig` and a `Metric.instance_key` property: the class name unless `name` overrides it.
- Route every keying site (`dispatch.py`, `optimize.py` instance dict / loss-log key / overlap check, `output.py`) through `instance_key`.

Default behavior is unchanged — `instance_key` falls back to the class name, which equals the existing `type` literal. Two tests cover the new coexistence and the still-rejected collision.

```yaml
pd.loss_metrics:
  - type: PGDReconLoss        # 1-step, auto-evaluated too
eval.metrics:
  - type: PGDReconLoss
    name: PGDReconLoss_20step  # distinct instance_key -> no collision
```

### DIFF
--- a/param_decomp/metrics/base.py
+++ b/param_decomp/metrics/base.py
@@ -63,8 +63,7 @@ class Metric[TConfig: BaseConfig](ABC):
         A loss-capable config may override it via `LossMetricConfig.name` so two
         instances of the same metric class (one loss, one eval) stay distinct.
         """
-        cfg = self.cfg
-        name = cfg.name if isinstance(cfg, LossMetricConfig) else None
+        name = self.cfg.name if isinstance(cfg, LossMetricConfig) else None
         return name if name is not None else type(self).__name__
 
     def bind(self, *, model: ComponentModel, device: str) -> None:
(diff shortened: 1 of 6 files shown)


param_decomp/metrics/base.py
```diff
@@ -50,6 +56,17 @@ def __init__(self, cfg: TConfig) -> None:
...
 
+    @property
+    def instance_key(self) -> str:
+        """Identity for dict keys and log-key suffixes; defaults to the class name.
+
+        A loss-capable config may override it via `LossMetricConfig.name` so two
+        instances of the same metric class (one loss, one eval) stay distinct.
+        """
+        cfg = self.cfg
+        name = cfg.name if isinstance(cfg, LossMetricConfig) else None
```
ocg-goodfire:
```suggestion
        name = self.cfg.name if isinstance(cfg, LossMetricConfig) else None
```

ocg-goodfire (approved):
LGTM 👍

Issue #729, opened 2026-06-16
Stage-4: support refused PD subspace variation points (delta off / tied / sigmoid_type / identity gating)


ocg-goodfire:
The JAX convert path hardwires the production PD subspace and refuses the rest at convert time: `sigmoid_type` must be `leaky_hard`, `use_delta_component` must be True, `tied_weights` must be None, `identity_decomposition_targets` must be None (E28, PARITY_MATRIX §11 row 11, status partial). These are real torch capabilities deliberately deferred (full-parity target), not na-by-design — part of the Stage-4 set.

jax ref verified: `param_decomp_jax/jax_single_pool/torch_config.py:85,232-233` (asserts). torch ref: `param_decomp_config/pd.py:164,170,147`.

Acceptance: each refused field, when supported, gets a JAX code path + an equivalence fixture against torch for that variation (e.g. delta-off forward, a tied-weights config, an alternate sigmoid_type, identity gating). Closeable per-field; this issue tracks the set. Cites S1, E28. NOTE: the persistent-PGD scope/sigmoid/start_frac/sign Stage-4 items live in the adversary area, not here — this issue is the config-gating model-data side only.

ocg-goodfire:
Ambiguous under the JAX-first pivot: Stage-4 deferred trainer feature (not a dropped feature, not a consumer port). Left open, no push label — revisit when the deferred PD-subspace variation points are scheduled. Under JAX-first it would default to matching torch (the oracle).

ocg-goodfire:
this is fine. can re add things when acutally used an otherwise it's good they're gone

PR #776, opened 2026-06-16
feat(jax): support pd.sigmoid_type variants in the CI fn (partial #729)

ocg-goodfire:
Partial close of #729 — implements the **`sigmoid_type`** variation point (one of the four refused PD-subspace fields). The other three (`use_delta_component=False`, `tied_weights`, `identity_decomposition_targets`) remain deferred and still refuse at convert time; #729 tracks the set and stays open.

## What changed
The JAX convert path hardwired `assert sigmoid_type == "leaky_hard"`. Now the torch `pd.sigmoid_type` threads through `CIArch` → `CIFn` (static field), and `ci_fn.squashing_fns(sigmoid_type)` selects the `(lower, upper)` squashing pair **exactly as torch `ComponentModel` does** (component_model.py:180-186):
- `leaky_hard` → asymmetric production pair (`lower_leaky_hard` custom-VJP + `upper_leaky_hard`)
- every other type → one function for both views

Added JAX impls of `normal` / `hard` / `leaky_hard` / `swish_hard` mirroring torch `param_decomp/ci_sigmoids.py`; `CIArch.sigmoid_type` defaults to `"leaky_hard"` (matches the torch schema default → existing fixtures unchanged); dropped the leaky_hard-only assert; SPEC variation-points CI-squashing row updated.

## Verification
- New `tests/test_ci_sigmoids.py`: torch-math parity for all 5 variant forwards, the `lower_leaky_hard` custom-VJP **backward** vs torch's `LowerLeakyHardSigmoidFunction.backward` (both grad signs), and the `squashing_fns` dispatch contract. Reference values are torch's exact closed forms encoded as numpy — no torch import needed.
- New `test_torch_config.py::test_sigmoid_type_threads_into_ci_arch`: a non-production `sigmoid_type` now converts (no longer refused) and flows verbatim.
- `python -m py_compile` clean; ruff lint+format passed. NOT run under JAX (deps-light worktree); pyright is a no-op here (param_decomp_jax is out of its include).

Equivalence here is pinned against torch's sigmoid math directly rather than the full fixture harness (which needs both a torch and a JAX env to regenerate goldens). Forward + backward are pinned for the custom-VJP variant.

### DIFF
--- a/param_decomp_jax/jax_single_pool/SPEC.md
+++ b/param_decomp_jax/jax_single_pool/SPEC.md
@@ -253,7 +253,7 @@ init: biases zero; weights fan-in scaled (torch init_param_)
 | `SCOPE` | persistent: `c (1,1)` · `sc (1,T)` · `nsc (n,T), n|B` · `bsc (B,T)` (jax: `sc` only today) — fresh-PGD: `c` · `bc` · `bsc` | ★ sc |
 | stoch sampling | `continuous U[0,1]` · `binomial {0,1}` | ★ continuous |
 | delta component | on (`C+1` channels) · off (no delta path, no delta mask) | ★ on |
-| CI squashing | `leaky_hard` pair (§4.2); other registry sigmoids exist in torch but are out of spec scope | ★ leaky_hard |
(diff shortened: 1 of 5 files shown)


ocg-goodfire:
Stale against the chunkwise refactor + config changes. #729 (PD-subspace variation points) stays open as Stage-4 backlog — redo cleanly when scheduled.

PR #805, opened 2026-06-16
feat(jax): persistent-PGD bsc scope (batch-sharded, no replica sync)

danbraunai-goodfire:
## What

Adds the `bsc` (`per_batch_per_position`) persistent-PGD source scope to the JAX single-pool trainer. Until now only `sc` was supported — a single `(1, T, C+1)` source shared across the whole global batch. `bsc` gives each batch element its own `(B, T, C+1)` source.

This is the last gap for replicating torch run **p-20f9fc15** — the same config José used for the paper's parameter decomposition (the "Interpreting Language Model Parameters" 4L Pile run, `param_decomp_lab/experiments/lm/pile_llama_simple_mlp-4L.yaml`). The `param_decomp_config` schema already supported `bsc` (incl. the `per_batch_per_position -> bsc` legacy alias); the gap was only in the JAX trainer.

## How bsc works

- `bsc` sources are **batch-sharded over `dp`** (`P("dp", None, None)`), so each element's source lands on the same rank as that element's `shard_batch`-placed residual/CI.
- Per-element grads are therefore **shard-local — no cross-rank source-grad reduction** (matching torch's `_skip_all_reduce`). The loss `1/B` normalization supplies the averaging. (`sc`, by contrast, is replicated `P()` and AVG-reduces its grad across shards.)
- Requires `global_batch % n_dev == 0`, the same divisibility `shard_batch` already needs.

## Why the bsc change is small

The per-step train logic is already shape-agnostic over the source leading dims (it operates on `{state_key: {site: array}}` via scan/grad/ascend), so the change is confined to **init + placement + one assertion**:

- `init_persistent_sources` gains a plain `batch_dim` int — the scope's leading axis (`1` for `sc`, `B` for `bsc`). Keeping it a plain int (not the scope enum) keeps the eager init decoupled from the loss-config types.
- `init_sources_sharded` gains `scope` / `global_batch` and `match`es `scope -> (batch_dim, placement)`.
- `recon._assert_supported_persistent` now admits `SCScope | BSCScope`.
- `train.TrainState.sources` is unchanged in code; only its docstring widened.

The `init_persistent_sources` signature change cascades to ~11 callers (experiments + tests), all `sc`, passing `batch_dim=1`.

## out_dir derived cluster-side (2nd commit)

The jax wrapper hardcoded an absolute `out_dir`, while the torch launcher already derives outputs from `PARAM_DECOMP_OUT_DIR` (`$DATA_MOUNT/.../runs`, the current cluster). `jax_launch` is now symmetric: `out_dir` is **optional** in the wrapper and **minted at submit** (stamped into the workspace copy alongside `run_id`), defaulting to `PARAM_DECOMP_OUT_DIR/runs`. A wrapper may still set `out_dir` to override (e.g. the llama8b wrappers' `jax_runs`, left as-is). Dropped the now-redundant `out_dir` from the two pile wrappers that used the default (`pile_ppgd_bsc`, `pile_pgd1`).

## Config

Ships the canonical reproduction pair only:
- `configs/pile_ppgd_bsc_from_torch.yaml` (JAX wrapper)
- `configs/torch/pile_llama_simple_mlp_4l_ppgd_bsc.yaml` (the `pgd1` sibling with the `PersistentPGDReconLoss`/`bsc` term swapped in)

## Validation

- `basedpyright jax_single_pool` and `basedpyright` (root, incl. `jax_launch.py`) — 0 errors.
- Full `jax_single_pool` suite — **65 passed** at both 1 and 4 simulated devices. Adds a `bsc` placement/shape block to `tests/test_sharding.py`.
- Functional check of the new wrapper validation: no-`out_dir` validates and derives the current-cluster path; explicit `out_dir` override accepted; missing/unknown keys rejected.
- Replication run **p-7847c8aa** (16 H100, dp=16) tracks torch p-20f9fc15 within ~0.1–0.4% on total loss over the first several hundred steps.

## Scope note

A separate CI-fn grad-clip feature lives alongside this in the experiment worktree; it's intentionally **not** included here (the base `bsc` config has `ci_fn grad_clip_norm: null`, so it doesn't need it). It can land as its own PR.

### DIFF
--- a/param_decomp_jax/jax_single_pool/CLAUDE.md
+++ b/param_decomp_jax/jax_single_pool/CLAUDE.md
@@ -4,9 +4,10 @@ Single-pool VPD trainer in JAX, **generic over vendored LM targets**. The semant
 source of truth is `SPEC.md` (normative pseudocode + numbered invariants, grounded in
 the stable torch `param_decomp` impl). See `README.md` for the file map.
 
-Open items: persistent-source scopes beyond `sc`, sigmoid parameterization,
+Open items: persistent-source scopes `c`/`nsc`, sigmoid parameterization,
 `start_frac>0`, and the hidden-acts seam are deliberately refused (LOSS_PARITY_DESIGN
-§6 stage 4); SPEC S24's two torch-parity quirks (PPGD warmup route-all, fresh-PGD
+§6 stage 4). `sc` and `bsc` are supported (`bsc` is batch-sharded, no replica sync).
+SPEC S24's two torch-parity quirks (PPGD warmup route-all, fresh-PGD
 single routing draw) are pinned pending a team decision; CI-fn numerics (gelu erf,
 rms eps) unify with torch at a run boundary.
 
--- a/param_decomp_jax/jax_single_pool/LOSS_PARITY_DESIGN.md
+++ b/param_decomp_jax/jax_single_pool/LOSS_PARITY_DESIGN.md
@@ -364,14 +364,14 @@ divisibility for `nsc`) becomes a converter assert.
 | `ImportanceMinimalityLoss` | **already-runnable** | S7–S9, D2; `p_anneal_final_p=None` (constant p) is a trivial assert-relax |
 | `ChunkwiseSubsetReconLoss` | **already-runnable** | the production stochastic term |
 | `StochasticReconSubsetLoss` (uniform_k) | **already-runnable** | converted today as 1-chunk plan |
-| `PersistentPGDReconLoss` (sc, Adam, clamp) | **already-runnable** | the production adversary |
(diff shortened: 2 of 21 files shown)


ocg-goodfire:
Taken over + landed via #836 onto post-refactor feature/jax — thanks @danbraunai-goodfire. Reconciled the persistent-config hunk against the chunkwise refactor (single PersistentPGDReconLossConfig; bsc slots into SCScope | BSCScope) and the restructured jax-wrapper key constants (out_dir author-overridable + submit-minted, alongside the new --comment flag). Full jax suite 148 passed at 1 and 4 simulated devices, goldens green.

PR #844, opened 2026-06-16
deps: upgrade to transformers v5 (load HF targets in native dtype)

danbraunai-goodfire:
## Description

Pin `transformers>=5,<6`. This is the alternative to #559 discussed in Slack: rather than
add a manual `dtype` field targeting old transformers, just move to v5, whose
`from_pretrained` defaults `dtype="auto"` and loads HF checkpoints in their **native
precision** instead of upcasting to fp32. For a bf16 model like Llama-3.1-8B that halves
the weight footprint (~16 GB/GPU), with no code change at the call site.

Verified empirically: a bf16 checkpoint reloads as bf16; fp32 checkpoints (gpt2) are
byte-identical to before.

## Issues found while upgrading

transformers v5 pulls **huggingface_hub 1.x** (0.36 → 1.19), which has two breaking
changes the repo hit. Both are fixed here:

1. **`huggingface_hub` moved `requests` → `httpx` and dropped `configure_http_backend`.**
   That import is at module load in `infra/hf_http.py` (the #557 Hub-retry shim), so the
   whole `experiments.lm.data` chain failed to import. Rewrote the shim onto the new
   `set_client_factory` API with a small retrying `httpx` transport — same policy as
   before (retry connect/read timeouts, network errors, 429/5xx with jittered backoff;
   idempotent methods only). Confirmed metadata calls (`HfApi.repo_info`, the call
   `datasets` makes at startup) still bypass hub's built-in `http_backoff`, so the shim
   is still needed.

2. **`PreTrainedTokenizerBase.decode` is now typed `str | list[str]`** (was `str`). Six
   basedpyright errors in `app/backend/app_tokenizer.py`. Routed all decode calls through
   a `_decode` helper that asserts `str` (a flat list of ids always decodes to one str).

Other usages checked and OK: `transformers.pytorch_utils.Conv1D` (core decomposition) is
unchanged; `AutoTokenizer` / model-class imports are unchanged; the pretrain Llama→custom
conversion uses canonical module names (`q_proj`, `embed_tokens`, …) stable across v4→v5
(not exercised here — needs real Llama weights).

## How Has This Been Tested?

- Full suite: **421 passed, 4 skipped** under v5.
- Slow gpt2 end-to-end decomposition (`--runslow`): passed.
- `basedpyright` + `ruff check/format`: clean.
- Native-dtype load confirmed empirically (bf16 preserved, fp32 byte-identical).

Behavioral: bf16/fp16-checkpoint targets now load at native precision instead of fp32
(same point as #559 — autocast already made matmuls bf16, so the frozen-target reference
shifts only slightly). fp32 checkpoints are unchanged.

Supersedes #559.

### DIFF
--- a/param_decomp_lab/app/backend/app_tokenizer.py
+++ b/param_decomp_lab/app/backend/app_tokenizer.py
@@ -62,8 +62,15 @@ class AppTokenizer:
     def encode(self, text: str) -> list[int]:
         return self._tok.encode(text, add_special_tokens=False)
 
+    def _decode(self, token_ids: list[int]) -> str:
+        # transformers>=5 types decode as `str | list[str]`; a flat list of ids always
+        # decodes to a single str.
+        text = self._tok.decode(token_ids, skip_special_tokens=False)
+        assert isinstance(text, str)
+        return text
+
     def decode(self, token_ids: list[int]) -> str:
-        return self._tok.decode(token_ids, skip_special_tokens=False)
+        return self._decode(token_ids)
 
     def get_spans(self, token_ids: list[int]) -> list[str]:
         """Decode token_ids into per-token display strings that concatenate to the full text.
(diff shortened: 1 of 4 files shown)


danbraunai-goodfire:

danbraunai-goodfire:
@claude can you fix the issues brought up in your last review?

danbraunai-goodfire:
Won't be relevant when we merge feature/jax.

Issue #828, opened 2026-06-16
Generic model I/O + pluggable recon loss (keep (B,T))


ocg-goodfire:
Make the JAX trainer generic over model **input**, **output**, and recon **loss** — keeping the `(B,T)` structure — so bio-style targets (dict inputs, tuple outputs, geometric losses) work.

## Fixed waist
The `[B,T,d]` residual stays the fixed waist. Everything load-bearing operates on `(B,T)` and stays exactly as-is: masking, routing (`recon.py`), source scopes + DP semantics (`adversary.py`), imp-min, the CI-fn (`ci_fn.py`), normalization. No batch-dim generalization.

## The three edges
1. **Input edge** — `prefix_residual_fn(prefix_weights, inputs)`: widen `inputs` from `token_ids` to `Any`. Data loader's native batch passes straight through.
2. **Output edge** — `clean_logits`/`masked_logits` on `DecomposedLM`: widen RETURN type to generic `Any`. Masking still happens on `[B,T,d]` residual internally. Field NAMES kept; rename deferred.
3. **Loss edge** — pluggable `recon_loss_fn(clean_output, masked_output) -> scalar`. DEFAULT is current `kl_per_position` (LM byte-for-byte unchanged).

## Forcing-function fixture
Tiny `DecomposedLM` impl with `[B,T,d]` residual but dict input, tuple output, non-KL recon loss + test driving recon machinery end-to-end.

## Strict bar
- Full JAX suite green at both device counts.
- basedpyright clean.
- New fixture test passes.
- Stacked-parity + equivalence goldens untouched.

ocg-goodfire:
Shipped to feature/jax in 66c09ccb7 (squash, single commit). Three edges widened: input (prefix_residual_fn -> Any), output (clean_logits/masked_logits -> Any), loss (DecomposedLM.recon_loss_fn, default kl_per_position). Synthetic dict-in/tuple-out/MSE fixture added (tests/test_generic_model_io.py). Full JAX suite green at both device counts (153 passed); stacked-parity + equivalence goldens byte-identical and passing.

PR #884, opened 2026-06-22
refactor(lm): DecomposedModel as Protocol + eqx.Module targets, + core-review cleanups

ocg-goodfire:
## Description

Stacked on #883. Lands the idiomatic-equinox `DecomposedModel` refactor plus the
core-review cleanups and a small reorg. Three buckets:

**1. `DecomposedModel` → Protocol + `eqx.Module` targets** (`7a79e1ea9`)
- The model interface is now a `@runtime_checkable Protocol`; concrete targets are
  `eqx.Module`s carrying their FROZEN target weights as array fields, with config as
  `eqx.field(static=True)`. The trainable V/U (`vu`) stays an explicit method arg
  (separate lifecycle: fp32 masters, own optimizer/checkpoint, C-sharded while frozen
  weights replicate).
- **No-bake invariant:** frozen weights reach the jitted step as a traced pytree ARG
  (`eqx.filter_jit`), never closed over — a closed-over 8B target would bake multi-GB
  into the HLO. Documented as law in `lm.py` and now **guarded by a test** (see below).

**2. Reorg**
- `param_decomp/targets/` — concrete target modules (`llama8b`, `llama_simple_mlp`,
  `target_aliases`, `llama8b_sharding`) moved out of the top level (`6591ac7d6`).
- `config.py` → `built_run.py` (`7021d28f9`).
- `log.py` library-safe: `NullHandler` at import, `setup_logger(logfile)` required and
  called from the run entry points — no handler/file config at import (`dab1cce87`).
- Pure shape primitives `SiteC`/`SiteSpec` moved `lm.py` → `components.py` (they belong
  with the V/U pytree they shape; resolves a real import cycle once `lm.py` depends on
  `DecompVU`).

**3. Core-review cleanups**
- Narrow `PersistentPGDReconLossConfig` `scope`→`SCScope|BSCScope`, `optimizer`→
  `AdamPGDConfig`; delete now-dead scope/optimizer types + a redundant runtime guard
  (`8bc85b7d6`).
- Move `runtime_cast` core→lab (its sole consumer is `harvest/reservoir.py`) (`79a5b1dee`).
- **Drop `run_id`/`out_dir` from the config schema** — they're minted identity, not
  authored spec. `run_id` is now a `--run-id` CLI arg (derived from the run-dir name on
  load); the run dir is `PARAM_DECOMP_OUT_DIR/runs/<run_id>` (single source). The dead
  `jax_runs` override (torch→JAX migration cruft) is gone (`011ca545e`).
- `components`/`vu` typed `DecompVU` instead of `Any` across the engine + Protocol; the
  one test that gratuitously rolled its own component type was rewritten to use the real
  `DecompVU` (the test bends to the clean core, not the reverse) (`e3ec35f88`).

## Related Issue
Addresses the core-code review triage and the generic-seam typing (#828).

## Motivation and Context
Idiomatic equinox modeling (weights as fields, not a function table), tighter types, and
a cleaner config/identity boundary, with the no-bake hazard turned from a comment into an
enforced invariant.

## How Has This Been Tested?
- `make check` → 0 errors/warnings/notes.
- Full suite @ 1 device **and** @ 4 sim-devices (`--xla_force_host_platform_device_count=4`):
  **448 passed, 9 skipped, 1 xfailed**, counts identical across both.
- Equivalence + stacked-parity goldens (`--runslow`): **16 passed, 1 xfailed**, unchanged
  (no `.npz`/`.npy`/golden data touched).
- New `test_no_bake_invariant.py` traces the real `make_train_step` jit boundary and
  asserts frozen weights are `jaxpr.invars`, not baked `consts` (verified it trips under
  the closure anti-pattern).

Yes — `run_id`/`out_dir` removed from the config schema. A before-validator strips both
from stored `config.yaml`s so existing runs still load; existing run dirs stay readable
(`open_jax_run` takes a run dir directly). Do not requeue an in-flight `jax_runs` run
across this change (its resume would look under `runs/`).


ocg-goodfire:
Landed in `feature/jax` via fast-forward (commits `7a79e1ea9..b4bda91f3` are now in the base). #883 (the CI-fn work it stacked on) is merged; this eqx + cleanups stack came along in the same FF. Closing — nothing left to merge.

PR #886, opened 2026-06-23
feat(eval): L0 + ci_masked-zeroed KL at multiple CI-aliveness thresholds

danbraunai-goodfire:
## Description

Generalize the CI_L0 eval from a single cutoff to a tuple **`ci_alive_thresholds`** (default `(0.0, 0.01)`):

- **L0 at each threshold** — `eval/l0/<thr>_<site|group>` scalars + one host-side bar chart per threshold.
- **`kl_ci_masked_zeroed_<thr>`** KL diagnostic at each *nonzero* threshold — the CI-masked forward but with every CI in `(0, thr]` dropped to exactly 0 (CI > thr kept). KL vs the plain `ci_masked` variant answers "do the tiny-CI components carry signal, or are they sediment?". It's a KL-only diagnostic, so it emits no `ce_difference_*` / `ce_unrecovered_*`.

The `0.01` cut matters because smooth-L0 (and L_p annealed toward 1) leave many components at small-but-nonzero CI; L0@0 counts them, L0@0.01 doesn't, and the `ci_masked_zeroed` KL shows whether dropping them changes the output.

## Motivation and Context

One clean concept, not two. The scalar `ci_alive_threshold` is replaced by the tuple `ci_alive_thresholds` **everywhere** (CI_L0Config, EvalConfig, `make_eval_step`, the LM eval builder, the bar-chart host code) — no parallel fields. The **first** entry is the primary threshold and still drives the slow-tier activation-density count. The four `param_decomp/configs/*` that set the old key are migrated; lab configs used the default and are untouched.

## How Has This Been Tested?

- `make type` / basedpyright clean (repo-wide).
- `test_eval.py`, `test_config.py`, `test_slow_eval.py`: 44 passed.
- All four migrated configs load via `LMExperimentConfig` (`ci_alive_thresholds == (0.0, 0.01)`).
- Exercised on btdr (2 nodes / 16 GPU): both `eval/l0/0.0_*` and `eval/l0/0.01_*` scalars + bar charts log, and `eval/ce_kl/kl_ci_masked_zeroed_0.01` tracks `kl_ci_masked` early on (the (0,0.01] components look like sediment so far).

Config schema: `CI_L0`'s `ci_alive_threshold: float` → `ci_alive_thresholds: tuple[float, ...]`. Run configs that set the old key must rename it (the in-repo ones are migrated here). `ComponentActivationDensity`'s own `ci_alive_threshold` is unchanged.

### DIFF
--- a/param_decomp/built_run.py
+++ b/param_decomp/built_run.py
@@ -69,7 +69,7 @@ class EvalConfig:
     `every` steps; the SLOW/plot tier (CI histograms, activation density, mean-CI,
     hidden-acts recon) runs every `slow_every` steps, reusing the same eval batches
     (SPEC S28/S29). `rounding_threshold` binarises CI for the CE/KL `rounded_masked`
-    variant; `ci_alive_threshold` is the CI-L0 aliveness cutoff."""
+    variant; `ci_alive_thresholds` are the CI-L0 aliveness cutoffs."""
 
     batch_size: int
     every: int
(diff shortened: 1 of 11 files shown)


danbraunai-goodfire:
Not sure if we need this for now. We will probably want it if we use smooth l0 AND if we find that smooth l0 has a lot of very small values that aren't useful.

PR #887, opened 2026-06-23
feat(losses): smooth-L0 (Geman–McClure) importance-minimality penalty

danbraunai-goodfire:
## Description

Adds `SmoothL0ImportanceMinimalityLoss` as a second importance-minimality penalty alongside the existing `L_p` one, ported from `feature/smooth-l0-investigation` onto the post-refactor feature/jax engine.

Per-value penalty `phi_gamma(c) = c² / (c² + gamma²)` — a smooth approximation to the active-component count `1[c>0]`, exact only as `gamma → 0`. It is flat at the origin (`phi'(0)=0`) and has a bounded gradient (`|phi'| ≤ 0.65/gamma`), so unlike `L_p` there is no singularity at `c=0` to `eps`-floor or aggressively grad-clip; the gradient is localized on the threshold band `c ~ gamma/√3` and redescends for clearly-on components. `gamma` anneals linearly (the `L_p` `p`-anneal analogue).

It feeds the **same** per-site `lp + beta · log2(1 + sum)` aggregation as `L_p` — the two penalties differ only in the per-value shape and which parameter anneals. So:
- `losses.py` factors the shared aggregation into `_imp_min_terms(penalty)` and the shared anneal into `_linear_anneal`; the trainer's single imp-min slot dispatches on config type via `annealed_imp_min_param` / `imp_min_terms`.
- The `L_p` path is **numerically unchanged** — same `importance_minimality_terms` / `annealed_pnorm` signatures, pinned by the existing equivalence + global-reduction tests.

## Motivation and Context

Enables a clean feature/jax run with the smooth-L0 sparsity loss (the loss your recent and-btdr runs used, which previously lived only on the smooth-l0 branch). Unblocks the chunkwise-CI 50k run.

## How Has This Been Tested?

- New `param_decomp/tests/test_smooth_l0_imp_min.py`: pins `phi` shape invariants (off→0, half-saturation `phi(gamma)=½`, clearly-on→~1), the flat-origin/bounded/redescending gradient, the per-site aggregation, and the anneal + dispatch.
- `param_decomp/tests/` green **except** two failures that are pre-existing on `feature/jax` and unrelated (verified by re-running on the base branch): `stacked_parity::test_clean_output_bit_identical` (≈1e-7 float reassociation vs a committed golden) and four `test_slow_eval` background-thread matplotlib renders — both local-environment.
- `make type` and `make format` clean; pre-commit passed.

No. Purely additive — a new opt-in loss type. No SPEC.md change (the docstrings cite `S7/S8/S9'`, mirroring the source branch, which shares S7/S8 with `L_p` and adds the `S9'` gamma-anneal analogue); flag if you'd prefer a formal SPEC entry.

### DIFF
--- a/param_decomp/configs.py
+++ b/param_decomp/configs.py
@@ -161,6 +161,38 @@ class ImportanceMinimalityLossConfig(LossMetricConfig):
     eps: NonNegativeFloat = 1e-12
 
 
+class SmoothL0ImportanceMinimalityLossConfig(LossMetricConfig):
+    """Geman–McClure smooth-L0 importance-minimality penalty on upper-leaky CI values.
+
+    Per-value penalty `phi_gamma(c) = c^2 / (c^2 + gamma^2)` — a smooth approximation to
+    the active-component count `1[c>0]`, exact only as `gamma -> 0` — fed through the same
+    per-site `lp + beta * mean * log2(1 + sum)` structure as `ImportanceMinimalityLoss`.
+    Differs from the `L_p` penalty only in the per-value shape: `phi'(0) = 0` and
+    `|phi'| <= 0.65/gamma` everywhere, so there is no singularity at the origin (no `eps`
+    floor, no aggressive grad clip) — the gradient is localized on the threshold band
+    `c ~ gamma/sqrt(3)` and redescends for clearly-on components.
+
+    `gamma` is annealed linearly toward `gamma_anneal_final_gamma` between
+    `gamma_anneal_start_frac` and `gamma_anneal_end_frac` of training; annealing it down
+    sharpens the count. A constant schedule is `gamma_anneal_final_gamma == gamma`.
+    """
+
+    type: Literal["SmoothL0ImportanceMinimalityLoss"] = "SmoothL0ImportanceMinimalityLoss"
+    gamma: PositiveFloat
+    beta: NonNegativeFloat
+    gamma_anneal_start_frac: Probability = 1.0
+    gamma_anneal_final_gamma: PositiveFloat | None = None
+    gamma_anneal_end_frac: Probability = 1.0
+
+
+# The two imp-min penalties share the `coeff`/`beta` surface and the `lp + beta * entropy`
+# aggregation; they differ only in the per-value penalty shape and its annealed parameter
+# (`p` vs `gamma`). The trainer's single imp-min slot accepts either.
+AnyImportanceMinimalityLossConfig = (
+    ImportanceMinimalityLossConfig | SmoothL0ImportanceMinimalityLossConfig
+)
+
+
 class CIMaskedReconLossConfig(LossMetricConfig):
     type: Literal["CIMaskedReconLoss"] = "CIMaskedReconLoss"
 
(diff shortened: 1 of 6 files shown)


danbraunai-goodfire:

param_decomp/configs.py
```diff
@@ -161,6 +161,38 @@ class ImportanceMinimalityLossConfig(LossMetricConfig):
...
+    gamma_anneal_final_gamma: PositiveFloat | None = None
+    gamma_anneal_end_frac: Probability = 1.0
+
+
+# The two imp-min penalties share the `coeff`/`beta` surface and the `lp + beta * entropy`
+# aggregation; they differ only in the per-value penalty shape and its annealed parameter
+# (`p` vs `gamma`). The trainer's single imp-min slot accepts either.
+AnyImportanceMinimalityLossConfig = (
+    ImportanceMinimalityLossConfig | SmoothL0ImportanceMinimalityLossConfig
+)
```
ocg-goodfire, line 193:
```suggestion
AnyImportanceMinimalityLossConfig = (
    LpImportanceMinimalityLossConfig | SmoothL0ImportanceMinimalityLossConfig
)
```

param_decomp/losses.py
```diff
@@ -54,20 +58,102 @@ def importance_minimality_terms(
...
+    total_steps: int,
+    initial: float,
+    final: float,
+    start_frac: float,
+    end_frac: float,
+) -> Array:
+    span = max(end_frac - start_frac, 1e-9)
+    progress = jnp.clip((step_f32 / total_steps - start_frac) / span, 0.0, 1.0)
+    return jnp.asarray(initial + (final - initial) * progress)
+
```
ocg-goodfire, line 101: could be a good time to use a proper schedule here instead of another instance of the old hacky inlined anneal (schedules.py or something, I think)

param_decomp/losses.py
```diff
@@ -54,20 +58,102 @@ def importance_minimality_terms(
...
+    ci_upper: dict[str, Float[Array, "*leading _"]],
+    cfg: AnyImportanceMinimalityLossConfig,
+    annealed_param: Array,
+) -> tuple[Float[Array, ""], Float[Array, ""]]:
+    """Dispatch `(lp, entropy)` on the imp-min penalty kind, given its annealed parameter."""
+    match cfg:
+        case ImportanceMinimalityLossConfig():
+            return importance_minimality_terms(ci_upper, annealed_param, cfg.eps)
+        case SmoothL0ImportanceMinimalityLossConfig():
+            return smooth_l0_importance_minimality_terms(ci_upper, annealed_param)
```
ocg-goodfire, line 156: the fact that these 2 are used in serial in 2 mutually exclusive ways is pretty odd to me. I feel like there's a nicer factorization

ocg-goodfire (approved):

ocg-goodfire:
Ported into `feature/jax` via #895 (commit `230f82e52`), adapted to the flat `LossTerm` structure that landed in #894 (the original diff here predates that flatten). Goldens bit-identical and your `test_smooth_l0_imp_min.py` is in. Thanks @danbraunai — this PR and the older #879 can be closed as landed/superseded whenever you like.

PR #898, opened 2026-06-25
fix(load_run): replicate HF prefix without cross-host allgather (dp>=64 OOM)

danbraunai-goodfire:
## Description
The GPU memory of this allocation did not scale with batch size — it scaled with the data-parallel process count, because of the path fixed here.

`build_target` placed the frozen HF Llama prefix (embedding + pre-suffix blocks) replicated via `jax.device_put(host_array, NamedSharding(mesh, P()))`. In multi-host JAX that path runs `multihost_utils.assert_equal` → `process_allgather(tiled=True)` to verify the hosts agree, which tiles the ~1 GB bf16 embedding to ~`process_count` GB. Replace it with `jax.make_array_from_callback`, which builds the replicated global array directly from each host's local copy — no cross-host allgather.

## Motivation and Context
At `dp=64` (8 nodes) the tiled allgather needed ~64 GB in a single allocation and OOM'd every rank during model build (`RESOURCE_EXHAUSTED: 62.62GiB`). It was tempting to blame the doubled train batch, but per-device batch was unchanged (dp doubled with it) — the allocation tracks process count, not batch. `dp=32` survived only because the tile was ~half the size. The prefix is read from the same cached safetensors on every host, so the equality check is pure overhead — dropping it is safe and removes the process-count memory scaling.

## How Has This Been Tested?
- `make type` clean.
- A `dp=64` (batch 128) Llama-3.1-8B layer-18 decomposition that OOM'd at build on this exact path now clears model load and trains; `dp=32` runs unaffected.

No. Replicated placement is unchanged; only the construction path differs (skips the cross-host equality allgather).


danbraunai-goodfire:
@claude could you review this?

PR #902, opened 2026-06-26
Add from-scratch decomposable LM to nano_param_decomp

ocg-goodfire:
## Summary

Trains an interpretable LM **from scratch** with parameter decomposition (Lucius's idea) — no target model. Built off the nano single-file impl (`run.py`), per Oli's request to rewrite cleanly rather than patch the main train loop.

New files:
- `nano_param_decomp/from_scratch.py` — the method (self-contained single file, mirrors `run.py`'s A–J structure)
- `nano_param_decomp/from_scratch_simplestories.py` — toy SimpleStories-2L entry point (random-init arch, weights discarded)

## What changed vs `run.py`

- **No target model.** `ComponentLinear` is now a *pure* UV decomposition `out = ((x@V)*mask)@U (+bias)` — no `W_target` buffer, no `mode` toggle, no weight-delta, no routing-to-target.
- **Train everything.** A fresh random-init `LlamaSimpleMLP` has its decomposed linears swapped for `ComponentLinear`; nothing is frozen. One AdamW over V/U + CI transformer + embeddings + norms (weight decay on dim≥2 params only).
- **CI-fn input = the decomposition's own activations.** `SPDModule.forward` runs the unmasked (all-ones-mask) component forward, caches each layer's input, feeds the CI transformer.
- **CE, not KL.** Stochastic-subset and PPGD recon losses both score next-token CE. Subset routing with no target: non-selected layers run at mask=1.0 (full components).
- **Training losses:** importance minimality (with the log-description frequency term) + stochastic-subset CE + adversarial-PGD CE. No faithfulness, no warmup, no KL.

## Review changes (latest)

- **Unmasked-NTP-CE is now eval-only**, not a training term. Stochastic-subset routing already trains non-selected layers at mask=1.0, so the ones-mask anchor is exercised partially without a dedicated term; reported as the eval-only `UnmaskedReconLoss`. We watch the eval to decide whether to add it back in a later iteration. (`coeff_unmasked` removed.)
- **Main LR back to the established low `5e-5`** (was 1e-3). Components (V/U) and the CI fn must move slowly so the persistent adversary can track them; the from-scratch scaffold (embeddings, norms) currently trains at the same low rate. *Open: whether to give the scaffold a separate higher LR since it doesn't interact with the adversary.*

## Test plan

- [x] `ruff` + `basedpyright` clean
- [x] CPU smoke (tiny random model, 6 steps + eval): losses compute, slow eval + PPGD eval run, unmasked CE ≈ ln(vocab) at init, PPGD CE > unmasked (adversarial working), `lr/main=5e-5`
- [ ] Toy SimpleStories-2L GPU run — Lee's check: reaches sane CE + sparse/sensible CIs (L0 + per-component CI figs)
- [x] DDP multi-GPU sanity (8-GPU toy: NCCL up, DDP wrap clean, 23.8M trainable; unmasked forward touches every param so no `find_unused_parameters` needed; PPGD `autograd.grad(loss, sources)` doesn't fire param hooks)

## Causal CI fix (next-token leak)

The CI transformer was inherited bidirectional (`is_causal=False`) from `run.py`. That's benign with a frozen target (recon is KL to the target's *own* position-t output, independent of t+1), but in the no-target/next-token-CE setting the label at position t **is** token t+1 — a bidirectional mask can attend to t+1 and route it into the position-t mask, so the masked forward predicts the next token for free. The overlay exposed this: `ce_ci_masked`=0.011 ≪ `ce_unmasked`=3.92 ≪ target val CE 2.24, which is impossible for honest masking (masking can only *remove* capacity).

Fix (commit `dda009cc4`): `CIAttention` self-attention is now **causal** (`is_causal=True`), so mask[t] depends only on tokens ≤ t. Relaunched as wandb `3460gqh3` (job 130613); the leaky run was `17jrbqbm`.

### DIFF
--- a/nano_param_decomp/from_scratch.py
+++ b/nano_param_decomp/from_scratch.py
@@ -7,11 +7,15 @@ and *all* parameters — components (V/U), the CI transformer, embeddings, and n
 are trained jointly. The reconstruction objective is next-token cross-entropy, not
 KL-vs-target.
 
-Losses:
+Training losses:
   - importance minimality (with the log-description frequency term)
-  - next-token CE on the unmasked forward (the LM-fitting anchor)
   - next-token CE under stochastic-subset masking (decomposability pressure)
   - next-token CE under adversarial (persistent PGD) masking
+
+The unmasked-forward next-token CE is *eval-only* (reported as `UnmaskedReconLoss`), not a
+training term: stochastic-subset routing already trains non-selected layers at mask=1.0, so
+the ones-mask anchor is exercised partially without a dedicated term. We watch the eval to
+decide whether to add it back.
 
 What's deliberately gone vs `run.py`: the target model, the weight-delta / spillover
 component, faithfulness loss + warmup, KL losses, and the target/component mode toggle.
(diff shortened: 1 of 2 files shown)


ocg-goodfire:
Discarding the nano from-scratch rewrite per Lucius. Reverting to opus/decomposable-from-scratch (commit b6e1bffd7), which produced p-671a9a1d — the implementation we want.

PR #908, opened 2026-06-30
feat(arith-eval): in-loop arithmetic CI + activation grid eval for the L18 MLP

lee-goodfire:
## Description

Adds an in-loop **arithmetic-grid eval** (`ArithmeticCIGrid`) for LM decompositions. On a fixed `a × b` modular-arithmetic prompt grid (`"<a>+<b>="`), it renders — per decomposed component, at the `=` answer position — an `a × b` heatmap of **causal importance** and of the **pre-mask activation `x@V`**, plus `n_alive` / `n_dropped` and recon / L0 / PGD on the arithmetic probe.

- `prestage_arithmetic.py` — offline generator of the grid artifact (one prompt/row, fixed length, single-token answers; carries `(a, b)` + `answer_id`). Loaded directly by the eval, not via the streaming `LMDataConfig` loader.
- `arithmetic_eval.py` — one fused step returning both CI and `x@V` at the answer position (batch axis kept as the grid); the active set (max CI > threshold) is selected once and shared by the scalars and the top-`top_k` figures.
- `llama8b.masked_component_activations` — new seam exposing `x@V` per live site, extending `_run_masked_forward`'s `collect` plumbing (sibling of `masked_site_outputs`, SPEC S31). LM-only; the eval narrows to it via the `ComponentActivationModel` protocol.
- Config plumbing (`ArithmeticCIGridConfig` → `ArithmeticEvalConfig`), the rank-0 background `ArithmeticGridRenderer`, wired through `_make_arithmetic_eval` into `_make_lm_eval_fn` on the slow-eval cadence; `configs/llama8b_l18_arith_add.yaml`.

## Related Issue

None.

## Motivation and Context

We want to see how well a decomposition reproduces the target model's modular-arithmetic mechanism — the L18 MLP addition neurons from Feucht et al. ("Arithmetic in the Wild"), which fire on periodic functions of the sum `a+b` (base-10 Fourier features, periods 2/5/10) and so show diagonal period-stripes on the `a × b` grid. The `down`-site activation grids are the target to recover. This ports the relevant slice of the arithmetic analysis from the torch `experiment/8B_targeted` work into the JAX line as a first-class eval, parameterised by site/threshold so it generalises beyond addition.

## How Has This Been Tested?

- Unit tests (CPU): `tests/test_arithmetic_eval.py` (fused step vs hand-rolled CI + `x@V`; row-major reshape; active-set selection; n_alive / n_dropped; render), `tests/test_llama8b.py` (the seam: `x@V @ U == masked_site_outputs` under all-ones masks), `experiments/lm/test_arithmetic_wiring.py` (probe-load order guard / missing-artifact / sharding). All green; `make type` (basedpyright) + ruff clean.
- GPU smoke: ran the eval at random init on the real Llama-3.1-8B L18 decomposition → CI + `x@V` heatmaps render to wandb; `n_alive ≈ 93%` at init, confirming the `top_k` cap and that n_alive starts high and falls as training sparsifies.
- Note: with the cuda jax extra installed, model-forward tests must run with `JAX_PLATFORMS=cpu` (cudnn flash rejects the tiny-model's fp32 attention on GPU).

No — additive. New optional eval metric; `masked_component_activations` is added only on `LlamaDecomposedModel`; `_run_masked_forward` gains an internal `collect_activations` arg (all call sites updated). Training / recon semantics are unchanged (the eval is figure-tier, off the recon path).


lee-goodfire:
Pushing a small fix. Ongoing tests surfaced an issue with the eval that depended on kernel selection, which showed up using some reasonable configs but not others. Fixing.

lee-goodfire:
> Pushing a small fix. Ongoing tests surfaced an issue with the eval that depended on kernel selection, which showed up using some reasonable configs but not others. Fixing.

Actually will wait until Oli's review is in. Don't want to break his work mid-flight.

PR #910, opened 2026-07-01
fix(runs): pin run config as launch_config.yaml, not config.yaml

danbraunai-goodfire:
## Description

Renames the pinned per-run config file from `runs/<id>/config.yaml` to `runs/<id>/launch_config.yaml`, routed through a single shared `LAUNCH_CONFIG_FILENAME` constant in `param_decomp/built_run.py` and used by every writer (LM + TMS + ResidMLP trainers) and reader.

## Motivation and Context

`config.yaml` is a **reserved filename in wandb's own run-files directory**. The trainer pins each run's self-contained config to `run_dir/config.yaml` (`_pin_config_copy`), then `MetricsSink.for_run` calls `wandb.save(run_dir/"config.yaml", base_path=run_dir, policy="now")`. That symlinks wandb's `files/config.yaml → run_dir/config.yaml`, and wandb writes its *own* run config through the symlink — clobbering the pinned launch config with a `_wandb:`/`<config>:` value dump.

Confirmed live on a finished run (cw-rno2, `p-1e7e8e36`): the run-dir `config.yaml` is a wandb dump, and `wandb/run-*/files/config.yaml` is a symlink back into the run dir.

In-loop readers survive (they read the config before `wandb.init`), but every **off-loop** consumer of a finished run's config reads the clobbered dump and fails:
- `load_run_dir_config` → `open_jax_run` / `run_metadata` / harvest / autointerp
- fine-tune structural-compat (`assert_finetune_structural_compat`)
- slow-eval metric re-validation (`eval_metrics_from_run_dir`)

A non-reserved basename removes the collision, so the pinned config stays clean and off-loop consumers work directly (no more "use the source config" workaround).

## How Has This Been Tested?

- `make type` — 0 errors
- `make format` / pre-commit — clean
- `pytest param_decomp/tests/test_config.py::test_load_run_dir_config_rebuilds_runs param_decomp/tests/test_finetune_resume.py param_decomp_lab/tests/autointerp/test_adapter_routing.py` — 8 passed
- grep confirms no lingering run-dir `config.yaml` reader/writer (autointerp *subrun* `config.yaml` is a different file, untouched)

Behavioral, for **old** runs only, and no shim is added (per repo policy). New runs write `launch_config.yaml` (clean). Existing finished runs only have the corrupted `config.yaml` dump — which the consumers already couldn't parse — so they now report as not-loadable rather than crashing mid-parse. To reopen a specific old run, drop its source YAML in as `launch_config.yaml`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,7 +202,7 @@ Every artifact for a decomposition lives under one dir per run:
 
 ```
 PARAM_DECOMP_OUT_DIR/runs/<run_id>/
-  config.yaml                # the single self-contained run config (the trainer reads it; resume byte-compares)
+  launch_config.yaml         # the single self-contained run config (the trainer reads it; resume byte-compares). NOT config.yaml: that basename collides with wandb's reserved run-config file, which wandb.save would symlink onto and clobber
   ckpts/<step>/...           # orbax sharded checkpoints (JAX trainer)
   metrics.jsonl              # local logs
   harvest/h-*/...            # pd-harvest output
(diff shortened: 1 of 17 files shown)


ocg-goodfire (approved):
TY!

PR #914, opened 2026-07-01
feat(logging): window min/max/median for grad-norm summaries

ocg-goodfire:
## What

The train loop **subsamples** metrics at the log step rather than reducing over the window since the last log. For most metrics that's fine, but it loses the shape of grad-norm spikes between logs.

For `grad_norms/summary/{components,ci_fns,total}` — the clip-relevant aggregates — this logs **min / max / median over every step since the last log**, replacing the single subsampled point value. Per-leaf grad norms and every other metric are unchanged (still subsampled).

New keys (via the sink's `train/` prefix): `train/grad_norms/summary/{components,ci_fns,total}/{min,max,median}`. They stay non-verbose, so the summary norms still print to the console.

## How

The per-step values are accumulated as **device handles**, not floats:

- Appending a handle each step is async — the loop stays unsynced between logs (which is load-bearing for throughput).
- At the log boundary the whole window stacks into a `[keys, steps]` array and reduces in a **single host transfer** — same one-sync-per-log cost as before, just over more values.

Three edits, all in `param_decomp/run.py`:
1. `_grad_norm_summary_window_stats` helper (stack + min/max/median).
2. Accumulate `grad_norms/summary/*` handles after each `step_fn` call.
3. At the log boundary, drop the instantaneous summary keys from the record and replace with the windowed stats; clear the window.

## Testing

- `basedpyright` + `ruff` clean (pre-commit passed).
- Helper math hand-verified on a known window.
- `stacked_parity` tests assert on the step's metrics dict (from `train.py`, unchanged), so they're unaffected.

https://claude.ai/code/session_01GjuHowooGKrpNaGCXmiztg

### DIFF
--- a/param_decomp/run.py
+++ b/param_decomp/run.py
@@ -156,6 +156,25 @@ def _is_verbose_grad_norm(key: str) -> bool:
     return key.startswith("train/grad_norms/") and not key.startswith("train/grad_norms/summary/")
 
 
+def _grad_norm_summary_window_stats(window: list[dict[str, jax.Array]]) -> dict[str, float]:
+    """Window min/max/median for each `grad_norms/summary/*` scalar over every step since the
+    last log. The per-step values are accumulated as device handles (appending one is async),
+    so the whole window reduces in a single host transfer here — the loop stays unsynced
+    between logs rather than subsampling grad norms at the log step."""
+    assert window, "grad-norm summary window is empty at a log boundary"
+    keys = list(window[0].keys())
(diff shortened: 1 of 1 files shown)


danbraunai-goodfire:

danbraunai-goodfire:
@claude I'm a bit concerned with this taking a long time to run and taking up a lot of memory. Could you evaluate both of these questions?

danbraunai-goodfire:
@claude could you update the PR with this suggestion "assert window, "grad-norm summary window is empty at a log boundary""

PR #922, opened 2026-07-02
Remove inert n_samples from PersistentPGDReconLoss

danbraunai-goodfire:
## Description

Removes the `n_samples` field from `PersistentPGDReconLossConfig` and pins the PPGD plan to `n_samples=1` in `build_loss_terms`. Stored run configs carrying `n_samples: 1` still load via the removed-field strip shim already on the class (the `use_sigmoid_parameterization` pattern); any other value is rejected at validation. The checked-in run yamls and the compile-probe generator drop the key.

## Motivation and Context

The PPGD term builds its plan with route-all routing and one persistent source bundle, so all `n_samples` routing draws produce byte-identical forwards — `n_samples > 1` just multiplied recon compute and averaged n identical losses. Sitting next to the stochastic losses where the same field genuinely means independent draws, it read as a variance-reduction knob and invited a silently wasteful sweep. It was `1` in every config; no setting of it was ever meaningful. (`LOSS_PARITY_DESIGN.md` had parked `n_samples>1` as an undecided composition-only variation point; this decides it. Plan construction is unchanged — SPEC S11/S24 untouched.)

## How Has This Been Tested?

- `make check` clean (ruff + basedpyright, whole workspace).
- `pytest param_decomp/tests -m "not slow"`: all pass except the `test_slow_eval.py` renderer tests, which fail identically on unmodified `feature/jax` (known flaky render-bound family).
- Extended `test_unsupported_settings_refuse`: stored-config `n_samples: 1` strips and loads; `n_samples: 2` rejects (same for both removed PPGD fields).
- Every tracked yaml in `param_decomp/configs/` validates through `LMExperimentConfig`.

A hand-authored config setting `n_samples != 1` on the PPGD term now fails validation — previously it silently burned compute. Stored run configs are unaffected (strip shim).

https://claude.ai/code/session_01KiMmzVhKUWbxbEz4R3Lkqm

### DIFF
--- a/param_decomp/configs.py
+++ b/param_decomp/configs.py
@@ -390,14 +390,20 @@ class PersistentPGDReconLossConfig(LossMetricConfig):
 
     @model_validator(mode="before")
     @classmethod
-    def _strip_removed_use_sigmoid_parameterization(cls, data: object) -> object:
-        # Shared-storage shim: stored run configs carry `use_sigmoid_parameterization`
-        # (always False — clamp was the only implemented path). The field is removed; strip
-        # it so those configs still load. A True value was never supported -> reject.
-        if isinstance(data, dict) and "use_sigmoid_parameterization" in data:
-            assert not data.pop("use_sigmoid_parameterization"), (
-                "use_sigmoid_parameterization was removed (clamp-only)"
-            )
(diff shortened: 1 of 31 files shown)


ocg-goodfire (approved):
TY!

Issue #237, opened 2025-10-28
Consolidate coefficient scheduling


danbraunai:
It'd be nice to have a standard way to manage schedules for lr, loss coefficients, and pnorm.

Oli has implement much of this [here](https://github.com/goodfire-ai/spd/tree/41de866b883a072d46757298105bace0630f222b).

ocg-goodfire:
Consolidation is done: unified `ScheduleConfig` (warmup + constant/linear/cosine) lives in `param_decomp/schedule.py` on canon (`feature/jax`) with tests, and lr/loss-coeff/pnorm route through it. The one remaining gap (p-annealing + audit for stray locally-rolled schedulers) is carved out into #915. Closing this per Oli.

PR #939, opened 2026-07-03
refactor(schedule): route p-annealing through unified ScheduleConfig + fold stray schedulers (S9/S13'/S20)

### LINKED ISSUE #915: Route p-annealing through unified ScheduleConfig (+ audit for stray locally-rolled schedulers)
Follow-up carved out of #237 (coefficient-schedule consolidation).

The unified `ScheduleConfig` (linear warmup + constant/linear/cosine decay) now lives in `param_decomp/schedule.py` (with `test_schedule.py`) on the canon branch `feature/jax`, and lr/loss-coeff/pnorm schedules route through it. The one remaining place that still uses a locally-rolled schedule is **p-annealing** (pnorm 2 → ~0.9).

We have a lot of config churn happening right now, so it's a good moment to finalise this:

- [ ] Route p-annealing through `ScheduleConfig` (`get_scheduled_value`).
- [ ] Audit the codebase for any other locally-rolled schedulers and fold them into `ScheduleConfig`.

Ref: #237. Canon branch: `feature/jax`.

ocg-goodfire:
# Route p-annealing through the unified ScheduleConfig (+ fold the stray schedulers) — closes #915

Follow-up to #237. Every scheduled quantity now has ONE config surface (`ScheduleConfig`) and ONE traced evaluator (`losses.scheduled_value_traced`, the jnp twin of the host `schedule.get_scheduled_value`, pinned pointwise by `test_schedule.py::TestTracedParity`).

## What changed

**1. p-anneal / gamma-anneal route through ScheduleConfig (S9/S9′).**
`ImportanceMinimalityLossConfig.pnorm` and `SmoothL0ImportanceMinimalityLossConfig.gamma` are now `ScheduleConfig`s; the windowed `{p,gamma}_anneal_{start,end}_frac` / `*_final_*` fields are deleted. Canonical `2.0 → 0.4` is now:

```yaml
pnorm:
  start_val: 2.0
  fn_type: linear
  final_val_frac: 0.2
```

- The window was dead expressiveness: every config in the repo AND all 33 stored `p-*` run configs use `start_frac=0.0 / end_frac=1.0`.
- Constant-`p` is `fn_type: constant` (kills the `final == initial` workaround and the `final_p is not None` asserts).
- Warmup on `p`/`gamma` is refused at `build_loss_terms` (a `p` ramping from 0 is never intended).
- **Trajectory change (deliberate, S9 amended):** the anneal now shares the schedule's `decay_steps − 1` denominator, reaching final `p` AT `step = steps − 1` instead of one step past the run (old `step / steps`). Per-step delta ≤ O(1/steps) — at 400k steps, ~4e-6 in `p`. No frozen goldens pin the old trajectory (the stacked-parity trajectory test is skipped; equivalence fixtures are per-term at a given `p`).

**2. Stray-scheduler audit — folded in (the issue's second checkbox):**
- `losses._linear_anneal` + `annealed_pnorm` + `annealed_gamma` → deleted (above).
- `losses.warmup_then_constant_lr` (PPGD source LR) → `scheduled_value_traced`. Trajectory identical for `warmup_pct > 0` (production uses 2.5%); at `warmup_pct == 0` the old JAX `max(warmup_steps, 1)` clamp gave source LR 0 at step 0 — that accepted seam vs torch is retired, JAX now matches torch (S13′ amended).
- `run_state.torch_cosine_schedule` (main VU/CI LRs) → `optax_schedule(cfg.lr_schedule, steps)` over the same traced fn. Previously only `start_val` was read and `alpha=0.1` was hardcoded (fn_type/final_val_frac/warmup_pct silently ignored — the config could lie about behavior). Trajectory bit-identical for the canonical cosine-to-0.1-no-warmup shape, which the lab conversion gate still pins (`assert_canonical_algorithm_config`). Torch parity re-pinned by `test_optim_torch_parity.py` (S20).
- **Not folded:** `pretrain/train.py::lr_schedule` — deliberately mirrors the torch pretrain recipe fields (`warmup_iters` absolute, `(it+1)/warm` ramp); folding would change pretrained-target reproducibility for no benefit.

**3. Layering kept:** the traced evaluator lives in `losses.py`, NOT `schedule.py`, so the config schema (`configs.py` → `schedule.py`) stays jax-free (its documented contract).

SPEC amendments (S9, S13′, S20) are in the diff, each tagged `AMENDED 2026-07-02 (#915)`.

## BREAKING: stored run configs need a one-shot migration

All 33 stored JAX runs' `launch_config.yaml` carry the old fields; `extra="forbid"` means resume / `load_run` / harvest of those runs fails until migrated. Their windows are all trivial, so the rewrite is mechanical. Run (after merge, once, against the runs dir):

```bash
python migrate_anneal_yaml.py $PARAM_DECOMP_OUT_DIR/runs/*/launch_config.yaml
```

(script attached in the PR conversation — same one that migrated the repo yamls; it refuses non-trivial windows and touches only the anneal lines, preserving formatting)

Note: resume byte-compares `launch_config.yaml` against itself, not against a re-render, so migrating the stored file is consistent for requeues; mid-flight runs on the OLD code must not be migrated until they move to this commit.

## Validation

- `pytest param_decomp/tests` at default device count and `XLA_FLAGS=--xla_force_host_platform_device_count=4`
- `tests/equivalence/` per-term goldens pass UNMODIFIED (they pin the imp-min term at a given `p`, not the anneal trajectory)
- `experiments/invariance_check.py` at 4 sim devices (D4)
- `pytest param_decomp_lab/experiments` (TMS / ResidMLP CPU trainers end-to-end through the new schema)
- `make check` (ruff + basedpyright) clean
- new `TestTracedParity` pins traced == host pointwise across fn_types × warmups × step counts; `test_optim_torch_parity` re-pins the torch cosine denominator


ocg-goodfire:
One-shot migration script for stored runs (referenced in the PR body). Run after merge:

```bash
python migrate_anneal_yaml.py $PARAM_DECOMP_OUT_DIR/runs/*/launch_config.yaml
```

```python
"""One-shot migration: windowed p/gamma anneal fields -> nested ScheduleConfig block.

Rewrites, per imp-min metric block:
    pnorm: 2.0                      pnorm:
    p_anneal_start_frac: 0.0    ->    start_val: 2.0
    p_anneal_final_p: 0.4             fn_type: linear
    p_anneal_end_frac: 1.0            final_val_frac: 0.2
(and the gamma_anneal_* analogue). Refuses non-trivial windows (start!=0 or end!=1).
Line surgery, not a yaml round-trip, so hand-authored formatting elsewhere survives.

Usage: python migrate_anneal_yaml.py FILE [FILE ...]
"""

import re
import sys
from pathlib import Path


def fmt(x: float) -> str:
    return repr(x)


def migrate(path: Path) -> bool:
    lines = path.read_text().splitlines(keepends=True)
    changed = False
    for value_key, prefix in (("pnorm", "p_anneal"), ("gamma", "gamma_anneal")):
        final_key = f"{prefix}_final_p" if value_key == "pnorm" else f"{prefix}_final_gamma"
        while True:
            group: dict[str, tuple[int, str, float]] = {}
            for i, line in enumerate(lines):
                m = re.match(
                    rf"^(\s*(?:- )?)({value_key}|{prefix}_start_frac|{final_key}|{prefix}_end_frac): ([0-9.eE+-]+)\s*$",
                    line,
                )
                if not m:
                    continue
                key = m.group(2)
                if key in group:
                    continue
                group[key] = (i, m.group(1), float(m.group(3)))
                if len(group) == 4:
                    break
            if len(group) < 4:
                assert len(group) == 0 or list(group) == [value_key], (
                    f"{path}: partial anneal group {sorted(group)}"
                )
                break
            indices = sorted(idx for idx, _, _ in group.values())
            assert indices[-1] - indices[0] <= 8, f"{path}: anneal keys too far apart"
            assert group[f"{prefix}_start_frac"][2] == 0.0, f"{path}: non-trivial start_frac"
            assert group[f"{prefix}_end_frac"][2] == 1.0, f"{path}: non-trivial end_frac"
            value_idx, lead, start_val = group[value_key]
            final_val = group[final_key][2]
            indent = re.sub(r"- $", "  ", lead)
            block = [f"{lead}{value_key}:\n", f"{indent}  start_val: {fmt(start_val)}\n"]
            if final_val == start_val:
                block.append(f"{indent}  fn_type: constant\n")
            else:
                block.append(f"{indent}  fn_type: linear\n")
                block.append(f"{indent}  final_val_frac: {fmt(final_val / start_val)}\n")
            for idx in sorted(indices, reverse=True):
                if idx == value_idx:
                    lines[idx : idx + 1] = block
                else:
                    del lines[idx]
            changed = True
    if changed:
        path.write_text("".join(lines))
    return changed


for arg in sys.argv[1:]:
    p = Path(arg)
    print(("migrated " if migrate(p) else "unchanged ") + str(p))
```

PR #950, opened 2026-07-06
feat(checkpoint): split checkpoints into decomposition | training items (S22/S33)

ocg-goodfire:
## What

Checkpoints store **two orbax items per step**, and `TrainState` COMPOSES exactly those two — one representation shared by the trainer and the checkpoint:

- **`decomposition`** — `train.Decomposition` (V/U `components` + `ci_fn`): the trained product.
- **`training`** — `train.TrainingItem` (both optimizer states, persistent adversaries, `step`): the process.

`TrainState = {decomposition, training}`, so `save_state`/`restore_step` map straight onto its own fields (no `_split`/`_join` regrouping). Every consumer — `open_jax_run` (harvest/autointerp/clustering/app) and fine-tune init (S33) — restores **only `decomposition`**, against an abstract tree from `jax.eval_shape(run_state.init_decomposition)`: zero knowledge of how training builds its optimizers or adversaries.

## Why

The 2026-07-06 harvest wedge: `open_jax_run` rebuilt + restored the FULL `TrainState`, so on one consumer GPU an 8B run's PPGD bsc sources + Adam moments (~60GB, sharded ÷64 during training) materialized whole, twice → OOM/allocator-wedge. The interim fix (1ca895860, only on local `harvest-tonight` — superseded here) partial-restored `{components, ci_fn}` but still needed `jax.eval_shape(init_train_state)` *including `build_optimizers`* just to recover the saved tree's structure. Consumers knowing training internals was the design smell; the item split removes it at the storage boundary.

## Compatibility — clean break, no in-code fallback

Clean checkpoint-format break. There is **no** dual-read of the pre-split single-`default`-item layout — every restore path reads the two-item layout only. Runs checkpointed before this change must be converted by an ad-hoc migration script; the compat logic does not live in the codebase. Both items save **sharded** — the no-on-loop-full-gather write property is unchanged.

## SPEC

S22 + S33 amended (marked **pending Oli sign-off** in the text): the two-item layout, the `TrainState`-composes-both invariant, the consumer contract, and the clean-break/no-fallback decision are normative.

## Validation

- Suite green at default device count and `XLA_FLAGS=--xla_force_host_platform_device_count=4`: `test_checkpoint`, `test_checkpoint_production_topology`, `test_generic_model_io`, `test_no_bake_invariant`, `test_llama8b`, `test_llama_simple_mlp`, `test_finetune_resume`, `stacked_parity`, and the lab `tms`/`resid_mlp` toy suites. `basedpyright` + `ruff` clean on the change. (The one failure, `test_pretrained_target_converts_with_wildcards`, is the pre-existing tip failure, unrelated.)
- Tests: two-item on-disk layout pin (cross-version contract), consumer host-restore vs saved state (bit-equal), two-item `init_from_parent` (S33: `decomposition` carries over; optimizer state / adversaries / `step=0` stay the fresh reference's), and `test_roundtrip_and_exact_resume` still bit-exact through the composed state.
- **New-format e2e smoke** (run on an earlier rev, persistent **bsc** adversary — the exact subtree that motivated the split): job 171284 trained 40 steps on 8 GPUs writing `<step>/{decomposition,training}`; job 171288 resumed from the two-item step-40 checkpoint; `open_jax_run(p-f653e1fd)` opened it in 6s on CPU and ran a clean harvest forward. Manual smoke config not committed.

## Notes for review

- `restore_step` keeps the format-coercion tail (`device_put` onto `reference.format`); #932 rewrites that same tail — whichever lands second has a small conflict.
- orbax gotcha in the tests: a `CheckpointManager` pins an item's handler at first use, so the saving manager can't `PyTreeRestore` what it `StandardSave`d — consumers open their own manager.
- Consequence of the clean break: pre-split runs (e.g. p-594db290) can't be opened by tip until migrated. The motivating harvest is separately blocked on `consumer-config-schema-drift`, so no live harvest regresses on this PR alone.

bridge thread: checkpoint-split-decomposition-item

Crew-Address: task/checkpoint-split-decomposition-item


param_decomp/checkpoint.py
```diff
@@ -33,6 +43,41 @@
...
+
+def _join(decomposition: Decomposition, training: TrainingItem) -> TrainState:
+    return TrainState(
+        components=decomposition.components,
+        ci_fn=decomposition.ci_fn,
+        components_opt_state=training.components_opt_state,
+        ci_fn_opt_state=training.ci_fn_opt_state,
+        adversaries=training.adversaries,
+        step=training.step,
+    )
```
ocg-goodfire: are these actaully necessary? they smell

PR #956, opened 2026-07-07
perf(compile): eliminate the init-compile share of the cold start

danbraunai-goodfire:
## Description

**Takeaways up front:**

1. **The "~20 min before step 0" was mostly init compiles, not the train-step compile.** Measured serially at the full production shape: `jit_init_decomp_vu` 211s + `jit_build_ci_fn` 167s + `jit_init_persistent_sources` 55s, before the ~5-min `jit_step` compile even started.
2. **This PR removes all three, bit-identically.** Same fix shape each time: the init was a jit emitting hundreds of separately-sharded outputs (or dozens of python-unrolled RNG bodies); it now generates vmap-stacked over the *same* per-site/per-chunk keys and fans out with a trivial donated slice jit. Values are exactly equal (pinned by `array_equal` tests), so nothing downstream changes — zero effect on step time, numerics, or checkpoints.
3. **Cold start at production scale drops ~13–17 min → ~6–8 min.** Verified end-to-end by rerunning last week's biggest btdr run (`p-283b7c88`, 64 GPUs): job start → training went ~17.6 min → ~12.1 min (the init share vanished; the rest is model load + `jit_step` + NCCL bring-up), with identical step time (4.3s), peak memory (68.4GB/rank), and loss trajectory.
4. **The remaining `jit_step` compile is not worth attacking structurally.** A 10-arm isolated-cache probe grid measured it flat across recon chunk count (1/2/4/8), halved C, CI-fn depth, and PPGD warmup (~5 min at dp32, +17% at dp64; ~83% XLA priority-fusion). The chunk-scan refactor scoped in #941 would buy nothing — documented in `param_decomp/CLAUDE.md` so nobody re-chases it.

Also included: a one-line `log.py` fix (`dictConfig` was disabling jax's already-created loggers, which is why `JAX_LOG_COMPILES=1` never appeared in any slurm log) and the measured findings + few-outputs-under-jit init practice in `param_decomp/CLAUDE.md`.

Supersedes #951 and #954 (their content, rebased onto the merged #955 and combined; V/U init originally from #941's bridge work).

## Related Issue

Supersedes #951/#954; continues #941's `reduce-jax-compilation-time` bridge task.

## Motivation and Context

Every new-config launch — smokes, sweep arms, debug iterations — paid ~7 min of pure init compile before anything useful happened. Sweeps paid it per arm.

## How Has This Been Tested?

- Bit-identity of all three inits pinned by exact-equality tests in `test_sharding` (1 and 4 sim devices); `make test` / `tests/equivalence` goldens / `make check` clean on this branch.
- GPU A/B at production shape (btdr, dp32, isolated caches): ci-fn 167→4.4s, sources 55→<2s, V/U 211→16s, `jit_step` unchanged.
- Max-scale end-to-end: exact rerun of `p-283b7c88` (full 32L, 64 GPUs, dp64) — trained 120 steps with identical step time/memory/losses before being cancelled.

No. Init values are bit-identical; the log change only re-enables foreign loggers.

### DIFF
--- a/param_decomp/CLAUDE.md
+++ b/param_decomp/CLAUDE.md
@@ -272,33 +272,17 @@ write cache entries from the first process … contention for writes on some fil
 so all ranks read but only rank 0 writes — no shared-FS race. Requires the cache dir on a
 shared FS, which `$PARAM_DECOMP_OUT_DIR` already is.
 
-### Cold-start compile budget (measured 2026-07-06, btdr, isolated-cache probe grid)
+### Compile time (measured 2026-07-06 probe grid; full data in PR #956)
 
-Method: one production-shape probe run per mutation (12 steps, eval/checkpoints off,
-`JAX_LOG_COMPILES=1`), each launched with its own `PARAM_DECOMP_OUT_DIR` so runs never
-share an XLA cache — cache isolation is a submit-time env var, nothing repo-side. Logs +
-launch configs archived on btdr (`compile_probe_grid_2026-07-06_logs.tgz`, sibling of
-`runs/`). Findings at the full 32L/224-site production shape:
-
-- The cold start is a SUM of serial compiles, and the inits used to dominate it:
-  `jit_init_decomp_vu` 211s + `jit_build_ci_fn` 167s + `jit_init_persistent_sources` 55s
-  before `jit_step` even started. All three are now vmap-stacked (few-outputs-under-jit;
-  `init_decomp_vu_placed` / `init_chunkwise_transformer_ci_fn` / `init_sources_sharded`),
-  bit-identically: measured after, V/U 16s, build_ci_fn 4.4s, sources <2s — the
-  historical "~24 min before step 0" was mostly this, not the step compile.
-- **`jit_step` compile is flat across GRAPH-STRUCTURE knobs**: recon chunk count 1/2/4/8,
-  halved C, 2-block CI fn, PPGD warmup on/off all land in the same ~4–6 min band at dp32
-  (PPGD removal saves ~1 min). Don't chase graph-shrink refactors for compile time
-  without new evidence.
-- **Topology slope is mild**: dp32 ≈ 290–306s, dp64 ≈ 345s (same code/hardware).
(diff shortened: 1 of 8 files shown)


danbraunai-goodfire:
Both review findings addressed in the latest commit:

**1. Inert donation — removed.** You're right: `donate_argnums=0` can never alias one stacked buffer into split per-site outputs, so the claim was false and every launch warned. The donation flags are gone from both fan-out jits, and the docstrings now state the true cost: init peak briefly holds one extra shard-local copy of the group stacks (÷N — a few GB/rank at production dp32, against an HBM that holds only the frozen target at that point in startup), freed when the fan-out returns. The compile-time win never depended on donation. Verified the warning is gone by running the sharding tests with `-W error::UserWarning` at 4 sim devices.

**2. Test gaps — covered.**
- `test_sharding`: `layers.{2,3}.mlp.gate_proj` now share `(d_in, d_out, C)`, so the V/U stacked init exercises a real multi-site shape group (asserted `max group size >= 2`), and the sources init gets a 3-site C group. The existing exact-equality pin vs the jitted per-site init now covers real stack/unstack.
- New `test_chunkwise_ci_init_vmap_matches_unrolled_reference` (ungated, plain `make test`): pins the vmapped CI init **bit-identical** (`array_equal`) against the unrolled+stacked per-chunk reference it replaced, at `n_chunks=3` with heterogeneous per-slot C — i.e. old-vs-new, not new-vs-new.

Full validation on the branch: sharding+llama tests at 1 and 4 sim devices (warnings-as-errors), equivalence goldens, `make check` — all clean.

ocg-goodfire:
oh shit this is sick

@claude is this the canonical jax fix here? idk why but something about having a vmap in the init feels weird to me. like, do we even need to compile the init? what if we just moved the compile boundary

ocg-goodfire:
@claude thanks.

can you explain this idiom?

```python
    stacked = jax.jit(
        partial(
            init_persistent_sources_stacked,
            site_names,
            site_component_counts,
            leading_shape,
            source_dtype,
        ),
        out_shardings=stacked_shardings,
    )(key)
```

ocg-goodfire (approved):
hell yea

PR #999, opened 2026-08-17
Consistent tPD metric names, and core evals on both streams

Antovigo:
## Description

* Enables standard evals (such as the PGD recon loss or L0) for **both of the tPD streams (target and nontarget)**
* Renames the tPD losses and evals so **the names are consistent.**

The keys are now `<tier>/[<stream>/]<family>/<name>`, where:
* tier: train or eval
* stream: `nontarget_data` for the nontarget stream in tPD, omitted otherwise
* family: ce_kl, l0, ...

## Motivation
In the current implementation of tPD, all standard evals run on the nontarget data stream. This isn’t explicitly labeled, and there’s no built-in way to run these standard evals on the target data instead. This PRs adds them, so any of the standard evals can be easily enabled or disabled on either of the streams.

There were also inconsistencies in the names of the metrics:
1. **Inconsistent position:** the stream segment sat *after* `loss/` on the train side (`train/loss/nontarget/<term>`), which only works because train has a single family. Eval has several (`ce_kl/`, `l0/`, `loss/`), so the segment has to come before them — both now put it directly after the tier.
2. **Multiple names for the imp-min losses.** The non-target imp-min *loss* was keyed by the step's short
   record key (`imp_smooth_l0`) while the target stream used the class name (`SmoothL0ImportanceMinimalityLoss`). Both streams now use the class name.
3. **Single-stream evals ran on the wrong stream:** attention-pattern recon, hidden-acts recon, `IdentityCIError`, `WellTemperedness`, the site figures (`CIHistograms`, `ComponentActivationDensity`, `CIMeanPerComponent`) and the permutation plots (`PermutedCIPlots`, `UVPlots`) now run on target data by default.
4. Introduces a **`WeightMagnitude`** plot, showing the components’ weight magnitudes ||U||*||V||, sorted by descending magnitude.

## What the metric names look like afterwards

**On a plain (non-targeted) run, nothing changes at all** — every key is what it is today.
The examples below are all from a **targeted** run:

| | before | after |
|---|---|---|
| non-target imp-min loss | `train/loss/nontarget/imp_smooth_l0` | `train/nontarget_data/loss/SmoothL0ImportanceMinimalityLoss` |
| its coefficient | `train/schedules/coeff/nontarget/impmin` | `train/nontarget_data/schedules/coeff/SmoothL0ImportanceMinimalityLoss` |
| non-target total | `train/loss/nontarget/total` | `train/nontarget_data/loss/total` |
| target-stream losses | `train/loss/total`, `train/loss/FaithfulnessLoss` | unchanged |
| L0 over nontarget data | `eval/l0/0.1_<site>` | `eval/nontarget_data/l0/0.1_<site>` |
| L0 over target data | *not measured* | `eval/l0/0.1_<site>` |
| 20-step PGD recon on target | *not measured* | `eval/loss/PGDReconLoss` |
| CI-masked KL on the target | *not measured* | `eval/ce_kl/kl_ci_masked` |
| well-temperedness, CI histograms, permuted-CI plots | `eval/slow/…`, `slow_eval/figures/…` over the **nontarget data** | same keys, now over the **target data** |

`CIMaskedReconLoss` and `UnmaskedNoDeltaReconLoss` become authorable under `eval.metrics`, like `PGDReconLossConfig`.
Authoring a narrow metric *and* `CEandKLLosses` is refused, because they would emit the same keys.

`TwoStreamCIMeanPerComponent` is like `CIMeanPerComponent`, except it shows both streams at once in two different colors.

## How Has This Been Tested?

- `basedpyright` clean; `ruff` clean (both also enforced by the pre-commit hook).
- Full non-slow suite over `core/tests`, `targets/tests`, `experiments`, `tests`:
  **1229 passed, 14 skipped, 11 xfailed**.
- Targeted TMS suite updated to the new train keys and passing.
- `test_eval_tier.py`'s authored-metric tier census updated for the four new eval-union
  members — it caught them, which is what it is for.

- The 4-simulated-device pass (`XLA_FLAGS=--xla_force_host_platform_device_count=4`) over
  `core/tests` + `targets/tests`, as `core/CLAUDE.md` requires: **593 passed, 12 skipped,
  11 xfailed**.

**Not for plain runs** — a plain run has one stream, so it never emits a stream segment and
every key is what it was.

**For targeted runs, yes**, in two ways. Every `nontarget/` key is renamed, so dashboards and
saved W&B views over tPD runs need updating and old/new runs will not overlay on one panel
for those series. And some evals that ran on nontarget data by default now run on target data.


param_decomp/core/run.py
```diff
@@ -333,7 +333,10 @@ def log(self, step: int, record: "LogRecord") -> None:
         self._last_committed_step = step
         record = {
             _METRIC_KEYS.get(
-                k, f"train/{k}" if k.startswith(("grad_norms/", "loss/", "schedules/")) else k
+                k,
+                f"train/{k}"
+                if k.startswith(("grad_norms/", "loss/", "schedules/", "nontarget_data/"))
```
Antovigo, line 338: The renamed non-target keys start with `nontarget_data/` rather than `loss/`, so without this entry they'd fall through to the `else` arm and be logged with no `train/` tier at all.

— *[written by Claude]*

param_decomp/core/configs.py
```diff
@@ -467,6 +468,15 @@ class ComponentActivationDensityConfig(BaseConfig):
     ci_alive_threshold: float = 0.0
 
 
+class WeightMagnitudeConfig(BaseConfig):
```
Antovigo, line 471: New plot. `slow: ClassVar[bool]` is required of every eval metric — `assert_every_metric_declares_its_tier` enforces that each class declares its own tier rather than inheriting one.

— *[written by Claude]*

Antovigo (commented):
Inline notes on the changes that aren't self-evident from the diff.

— *[written by Claude]*

param_decomp/core/slow_eval.py
```diff
@@ -198,6 +199,11 @@ def slow_eval_step(
     return filter_jit(slow_eval_step, compiler_options=compiler_options)
 
 
+def _raw_sample(chunks: dict[str, list[np.ndarray]], site: str) -> np.ndarray:
```
Antovigo, line 202: **Bug fix, not a refactor.** `n_batches_accum=0` means 'keep no raw sample', but the constructor used to `np.concatenate(chunks[site])` unconditionally — with no chunks that's a `KeyError`. `TwoStreamCIMeanPerComponent` passes 0, so it crashed on every invocation before this PR and has never actually run.

— *[written by Claude]*

param_decomp/core/slow_eval.py
```diff
@@ -578,6 +584,89 @@ def _plot_ci_matrices(matrices: dict[str, np.ndarray], colormap: str, title_pref
...
+        fig = Figure(figsize=(8 * n_cols, 3 * n_rows))
+        axs = fig.subplots(n_rows, n_cols, squeeze=False)
+        flat_axes = axs.T.ravel()
+        for ax in flat_axes[len(ordered) :]:
+            ax.set_visible(False)
+        for ax, (name, (target, nontarget)) in zip(flat_axes, ordered.items(), strict=False):
+            x = np.arange(len(target))
+            if log_y:
+                ax.set_yscale("log")
+            ax.fill_between(x, target, step="mid", color="#1f77b4", label="target")
```
Antovigo, line 657: `ax.bar` here creates one matplotlib patch per component: measured ~120s per figure at C=1456 over 32 sites, and it renders two figures on the background thread that competes with the train loop for the GIL. `fill_between(step='mid')` draws the same picture in ~23s for both.

— *[written by Claude]*

param_decomp/core/slow_eval.py
```diff
@@ -578,6 +584,89 @@ def _plot_ci_matrices(matrices: dict[str, np.ndarray], colormap: str, title_pref
...
+    """Per-site `‖V_c‖·‖U_c‖` in descending magnitude order, log y. x is a component's rank
+    within its site, NOT its component id."""
+    n_rows, n_cols = _grid_dims(len(magnitudes))
+    fig = Figure(figsize=(8 * n_cols, 3 * n_rows))
+    axs = fig.subplots(n_rows, n_cols, squeeze=False)
+    flat_axes = axs.T.ravel()
+    for ax in flat_axes[len(magnitudes) :]:
+        ax.set_visible(False)
+    for ax, (name, values) in zip(flat_axes, magnitudes.items(), strict=False):
+        ax.scatter(range(len(values)), np.sort(values)[::-1], marker="x", s=10)
```
Antovigo, line 619: Sorted descending, so x is a component's **rank within its site, not its component id** — you can no longer cross-reference a point against other per-component plots.

— *[written by Claude]*

param_decomp/core/train.py
```diff
@@ -1185,14 +1185,18 @@ def make_targeted_train_step[PreparedT](
...
-        "nontarget/impmin": objective.nontarget.impmin_coeff,
-        **{f"nontarget/{term.name}": term.coeff for term in nt_terms},
     }
     if objective.target.imp.cfg.frequency is not None:
         coeff_schedules[f"{objective.target.imp.name}/frequency"] = (
             objective.target.imp.cfg.frequency.coeff
         )
 
+    imp_name = objective.target.imp.name
+    nontarget_coeff_schedules: dict[str, LossCoeff] = {
```
Antovigo, line 1195: Split out so the non-target coefficients can take the `nontarget_data/` prefix at the one call site that needs it, leaving `_scheduled_coeff_metrics` exactly as upstream has it (the plain step's call is unchanged).

— *[written by Claude]*

param_decomp/experiments/lm/eval.py
```diff
@@ -204,6 +205,49 @@ def _ce[PreparedT](batch: _PreparedLMBatch[PreparedT], logits: Array) -> Array:
     return _row_masked_cross_entropy(logits, batch.tokens, batch.valid_row_mask)
 
 
+type MaskingArm = Literal["ci_masked", "unmasked"]
+"""A masking arm authorable as an eval on its own. Both pin every weight-delta mask to
+zero."""
+
+
+def make_masked_kl_step[PreparedT](
```
Antovigo, line 213: Deliberately a separate step rather than a `variants` parameter on `make_ce_kl_step`: the point of these single-arm evals is that a tPD run does *not* author `CEandKLLosses`, so teaching that shared bundle to select arms would couple this change to the thing it lets runs avoid. `make_ce_kl_step` is byte-identical to upstream.

— *[written by Claude]*

param_decomp/experiments/lm/scalar_eval_operations.py
```diff
@@ -13,15 +15,41 @@
...
     make_ce_kl_step,
     make_ci_l0_step,
     make_fresh_pgd_step,
+    make_masked_kl_step,
 )
 from param_decomp.experiments.lm.eval_config import CEandKLLossesConfig
 from param_decomp.experiments.lm.eval_context import LMEvalContext
 from param_decomp.experiments.lm.eval_keys import EvalKeyStream
 
+type Stream = Literal["nontarget", "target"]
```
Antovigo, line 29: One value, not a `(batches, log prefix)` pair — `stream_batches` and `stream_log_prefix` both take it, so an operation cannot read one stream's batches while writing the other's keys.

— *[written by Claude]*

param_decomp/experiments/lm/eval_context.py
```diff
@@ -11,3 +11,7 @@
 class LMEvalContext(EvalInvocation):
     pass_index: int
     batches: tuple[jax.Array, ...]
+    target_batches: tuple[jax.Array, ...] | None = None
```
Antovigo, line 14: Defaults to `None` so the plain LM root and its tests need no change at all — a plain run has one stream and says so by omission. That `None` is also what every log key reads to tell which run kind it's in.

— *[written by Claude]*

param_decomp/experiments/lm/eval_operations.py
```diff
@@ -87,102 +102,158 @@ def batches(pass_index: int) -> list[jax.Array]:
...
     def well_temperedness_inputs(
         context: LMEvalContext,
     ) -> tuple[jax.Array, PRNGKeyArray]:
-        return context.batches[0], jax.random.fold_in(
+        return stream_batches(primary_stream, context)[0], jax.random.fold_in(
             run_key, EvalKeyStream.WELL_TEMPEREDNESS * pd.steps + context.pass_index
         )
 
-    def make_operation(metric: AnyEvalMetricConfig) -> EvalOperation[LMEvalContext]:
+    def per_stream(
```
Antovigo, line 116: `make_operations` returns a tuple because one authored metric now yields one operation *per stream*. `all_streams` is both on tPD and just the one on a plain run; `primary_stream` is where single-stream evals bind (target on tPD, the only stream otherwise).

— *[written by Claude]*

param_decomp/experiments/lm/eval_operations.py
```diff
@@ -87,102 +102,158 @@ def batches(pass_index: int) -> list[jax.Array]:
...
 
-    operations = tuple(make_operation(metric) for metric in eval.metrics)
+    authored = {type(metric) for metric in eval.metrics}
+    assert targeted or TwoStreamCIMeanPerComponentConfig not in authored, (
+        "TwoStreamCIMeanPerComponent measures the target stream; a plain run has none"
+    )
+    assert not (
+        authored & {CIMaskedReconLossConfig, UnmaskedNoDeltaReconLossConfig}
+        and CEandKLLossesConfig in authored
+    ), "the single-arm KL evals emit keys CEandKLLosses also emits; author one or the other"
```
Antovigo, line 253: Bind-time rather than at the first eval pass: the operations would otherwise collide on identical `ce_kl/kl_*` keys and trip `_run_due_evaluation`'s assert hours into a run.

— *[written by Claude]*

param_decomp/experiments/lm/training_targeted.py
```diff
@@ -136,8 +139,27 @@ def sample_nontarget_batch(step: int) -> jax.Array:
             "eval must land on a train-log step: the tok/s window resets after eval, so a "
             "mid-window eval would corrupt the next step-time estimate"
         )
+        eval_target_batch = eval_config.batch_size
+
+        def eval_target_pool_batches(pass_index: int) -> list[jax.Array]:
```
Antovigo, line 144: The eval pass needs target-stream batches that `data.eval` can't supply. Uses training's own pure `(seed, step)` pool sampler on the `seed + 1` stream — the same stream the corpus eval split draws from — so an eval never scores the rows the step just trained on.

— *[written by Claude]*

